PR #5, opened 2021-04-21
Migrating to formik; plus conditional fields + field validation

nathandavies91:
Migrating forms over to [formik](https://formik.org/):

- Form/field level validation, with validation schemas built using [Yup](https://github.com/jquense/yup)... See `FormField.validation?`
- Ease of working with form values `values.{property}` so we can implement conditional logic around form fields... See `FormField.conditionalDisplay?`
- Plugged in the 'dynamic form' data to render forms using formik `<Field />` elements
- Implemented custom error handling using formiks validation logic

Additionally:

- Updated SASS version, just happen to be in npm at the time
- Tided up general approach to ensure a bit more consistency within the application

Roadmap:

Now that we have formik, it would be good to migrate the multi-step logic over also with; with a little help from these tutorials:

- https://codesandbox.io/s/on6vx42rq?file=/Form.js
- https://www.youtube.com/watch?v=kKpoAhugjUc


components/button.tsx
```diff
@@ -2,15 +2,24 @@ import Link from "next/link"
...
+export default function Button({ children, href, type }: ButtonProps): JSX.Element {
+  const button = (
+    <button className="govuk-button lbh-button" type={type}>
+      {children}
+    </button>
   )
+
+  if (href) {
+    return (
+      <Link href={href}>
```
tcmorris, line 18: This feels like an odd use of button, since it will only ever act as an `a` tag and not a button?

nathandavies91: Unfortunately, you cannot wrap the custom `<Button />` component with `<Link />`, for some reason that did not work - it's like it expects the component beneath it to be a button or anchor tag, so I had to move the link functionality into the button component.

I think this is fine, because in terms of display it appears as a button

tcmorris:
Appears to work ok to me, tested a few options. Happy to merge 👍

tcmorris (approved):

PR #7, opened 2021-04-23
Multi step forms

nathandavies91:
Reworking if a little bit so that we now have actual steps within the form.

Each step is using formik, and we pass it only the relevant validationSchema for that step; meaning future steps do not fail validation at this time. This also enables us to ensure everything is valid within the current step before the user can progress.

The submit and next functions are also called, so we can plug into the form and save the data as the user progresses through.


tcmorris (approved):
Looks good. Had a couple of thoughts, but these appear to be covered with the associated TODOs.

PR #11, opened 2021-05-10
HRT-25 & HRT-28: Eligibility checker

nathandavies91:
Quite a big overhaul of the application forms and eligibility checker:

Overview:

- Updating the form journey to closely match the Figma journey (agreement, overview of people [currently hardcoded], overview of tasks [test data], and the ability to progress through a complete application)
- Storing user state and form data within Redux; wrapping the nextjs app with redux wrapper for forcing state changes across the entire app, `useStore` and `useSelector` are great ways to request the store -- examples have been set-up in the app as is
- Breadcrumbs on parts of the app, to allow the user to easily navigate back when filling out an application (need to warn that progress will be lost)
- Conditional fields were a thing before, however, conditional validation needed setting up in line with the display state to ensure we don't block form progression when a hidden field is also 'required'

Eligibility checker:

- Optional "eligibility-checker" property as part of the form data
- On submitting the `ApplicationForm` component, the data is checked against the criteria set-up above
- If the data is deemed valid ("eligible") then the user state is updated as eligible and passed on
- If the data is seemed invalid ("ineligible"), then the user state is updated as ineligible and reasons are also recorded for display later in the app (this also helps when returning to the app, the state is remember and we can continue to show reasons why)
- A new `<EligibilityOutcome />` component has been created which checks the state set before and displays the relevant page in order to continue, either the Eligible page or the Ineligible page which details next steps and ways Hackney can support that person

Redux:

- Currently limited to just the `user` store, with a few properties and actions (like logIn, logOut, updateFormData, ...)... See `lib/store/user.ts`
-- User store currently holds the 'formData', this is because the data currently belongs to the user; with additional members also having formData of their own -- so it was just a potentially easy way to manage this. However, this might be a bad idea because when updating form data we are refreshing the user model and forcing logIn/logOut state to push potential updates also? Might be expensive for performance?


components/header.tsx
```diff
@@ -19,14 +22,16 @@ export default function Header({ userName }: HeaderProps): JSX.Element {
               </a>
             </Link>
           </h1>
-          {userName && (
+          {user.loggedIn ? (
```
tcmorris, line 25: We'll probably need to move the header in to each respective layout (resident/staff) since we have conflicting use cases here for what a user is.

pages/apply/[person]/[[...step]].tsx
```diff
@@ -0,0 +1,38 @@
...
+import Layout from "../../../components/layout/resident-layout"
+import { useRouter } from 'next/router'
+
+export default function ApplicationStep(): JSX.Element {
+  const router = useRouter()
+  let { person, step } = router.query
+  person = person as string
+
+  const baseHref = `/apply/${person}`
+  const steps: string[] = ["test-data", "immigration-status"]
```
Comment (tcmorris) [inline pages/apply/[person]/[[...step]].tsx:11]: Define via props?

pages/applications/[id]/index.tsx
```diff
@@ -2,8 +2,8 @@ import Layout from '../../../components/layout/staff-layout';
 import { GetServerSideProps } from 'next'
 import { getApplication } from "../../../lib/gateways/applications-api"
 import { Application } from '../../../domain/application';
-import { HeadingOne, HeadingTwo } from "../../../components/headings"
-import Paragraph from "../../../components/paragraph"
+import { HeadingOne, HeadingTwo } from "../../../components/content/headings"
+import Paragraph from "../../../components/content/paragraph"
```
Comment (tcmorris) [inline pages/applications/[id]/index.tsx:6]: Usage of `Paragraph` has changed here.

tcmorris (approved):
In theory, looks good. There's quite a lot going on here, but appears to be working ok. My knowledge of redux etc is minimal so could do with another pair of eyes, but happy to merge and update later.

pages/applications/[id]/index.tsx
Comment (nathandavies91) [reply pages/applications/[id]/index.tsx:6]: Sorted

pages/apply/[person]/[[...step]].tsx
Comment (nathandavies91) [reply pages/apply/[person]/[[...step]].tsx:11]: This is the entry point (the page), so we pass it through to `<ApplicationForms />`

PR #14, opened 2021-05-12
Prevent residents from accessing the application if they are ineligible

nathandavies91:
A new high-order component that restricts access to the application (`/apply`) if the use is ineligible for housing.

Any ineligible requests will be shown the `<EligibilityOutcome />` component which details further information and reasonings behind this decision (as stored against the resident state).

Undefined eligibility state means that the user has not taken the eligibility checker tool or we have no record of it, so we cannot currently determine if they are eligible or not - we should continue to allow them to apply, and the decision will be made as they progress through the application.

### DIFF
--- a/lib/hoc/whenEligible.tsx
+++ b/lib/hoc/whenEligible.tsx
@@ -1,0 +1,26 @@
+import EligibilityOutcome from "../../components/eligibility"
+import Layout from "../../components/layout/resident-layout"
+import { Store } from "../store"
+import { MainResident } from "../types/resident"
+import React from "react"
+import { connect } from "react-redux"
+import { compose } from "redux"
+
+const whenEligible = (WrappedComponent: React.ComponentType<any>) => {
+  return (props: MainResident) => {
+    if (props.isEligible === false) {
+      return (
+        <Layout>
+          <EligibilityOutcome />
+        </Layout>
+      )
+    }
+
+    return <WrappedComponent {...props} />
+  }
+}
+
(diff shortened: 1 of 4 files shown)


tcmorris:
From initial testing it looks like you can do the eligibility steps and then go straight to `/apply/overview` and see the page as normal?

nathandavies91:
> From initial testing it looks like you can do the eligibility steps, get a confirmed ineligible outcome and then go straight to `/apply/overview` and see the page as normal? which I wouldn't expect to work.

@tcmorris If you are doing this in the browser and not using the `<Link />` elements (like clicking the logo and hitting start), then it's likely you are starting a new session and clearing the resident store (this is not stored locally or against an API just yet).

Apologise, I should have included test steps:

1. Fill out eligibililty checker tool ('/eligibility-checker')
2. Get an ineligible response by selecting 'European' and 'Studying in the UK'
3. Click the logo to go back to the home page (without wiping the resident store)
4. Then click 'start' ('/apply'), this should give you the ineligible screen and not the agreement screen... If you refresh you should see the agreement string (as session has been cleared and you're able to view the application as a decision has not been made at this point)

1. Re-try with an eligible response (by selecting 'British' resident), and re-try -- this time you should be able to click through to the application form and be met with the agreement page (not blocked at all)

tcmorris:
Ah ok, did not realise that was the case. Your provided example works as expected.

tcmorris (approved):

PR #15, opened 2021-05-12
Forms: set focus to the first field

nathandavies91:
A simple fix which jumps the user to the top of the form when they progress onto the next/previous step of a multi-step form.

Previously this was keeping the users' focus at the bottom of the new step, which is not idea.

### DIFF
--- a/components/form/checkboxes.tsx
+++ b/components/form/checkboxes.tsx
@@ -10,7 +10,7 @@ interface CheckboxProps extends FormField {
   value: string
 }
 
-export function Checkbox({ hint, index, label, name, value }: CheckboxProps): JSX.Element {
+export function Checkbox({ focus, hint, index, label, name, value }: CheckboxProps): JSX.Element {
   let id = name
 
   if (index) {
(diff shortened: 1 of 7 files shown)


tcmorris:
Noticed a couple of things from testing:
- focus is not assigned to validation errors
- focus on radio button causes a validation error before a selection is made (e.g. What nationality are you?)

nathandavies91:
> Noticed a couple of things from testing:
> 
> * focus is not assigned to validation errors
> * focus on radio button causes a validation error before a selection is made (e.g. What nationality are you?)

Fair point.

- We could do it so that they are scrolled to the top of the page always, but that could also miss the form if it sits down the page a fair way (like on mobile).
- As for validation, this is to do with progression of steps. In the event of validation it should scroll back up, although I have noticed that the first field for some reason does not work (why, I don't know, a formik issue here?)... The second field (like email validation) always seemed to work though

tcmorris:
In that case, I'm not sure we want the focus to be the first input on load since it's plausible that you'd miss information above the form? e.g. introductory text or supporting information about what is required? In this scenario, the top of the page (or default behaviour of the page url) makes sense.

nathandavies91:
> In that case, I'm not sure we want the focus to be the first input on load since it's plausible that you'd miss information above the form? e.g. introductory text or supporting information about what is required? In this scenario, the top of the page (or default behaviour of the page url) makes sense.

@tcmorris Ok, lets leave this for now and revisit it later..

PR #16, opened 2021-05-12
HRT-48: Application start screen, adding example wait times

nathandavies91:
Superficial changes to the start page:

1. Amending copy to match latest figma designs
2. Adding additional copy
3. Adding example of wait times (not currently driven by any data)
4. Wrapping wait times in the standard `<Accountment />` component (`info` variant is being used) - this does not currently match Figma, but we should probably stick to the design system here

### DIFF
--- a/components/announcement.tsx
+++ b/components/announcement.tsx
@@ -1,0 +1,18 @@
+interface AnnouncementProps {
+  children: JSX.Element | JSX.Element[]
+  variant: "info" | "success" | "warning"
+}
+
+export default function Announcement({ children, variant }: AnnouncementProps): JSX.Element {
+  let className = "lbh-page-announcement"
+
+  if (variant) {
+    className += ` ${className}--${variant}`
+  }
+
+  return (
+    <section className={className}>
+      {children}
+    </section>
+  )
+}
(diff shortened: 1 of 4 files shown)


components/announcement.tsx
```diff
@@ -0,0 +1,18 @@
...
+export default function Announcement({ children, variant }: AnnouncementProps): JSX.Element {
+  let className = "lbh-page-announcement"
+
+  if (variant) {
+    className += ` ${className}--${variant}`
+  }
+
+  return (
+    <section className={className}>
+      {children}
```
tcmorris, line 15: This doesn't appear to tally up with the design system, note the lack of `title` and `content` which I think would be required if we want to keep it as a `section` where a heading should be provided.
https://design-system.hackney.gov.uk/components/page-announcement

components/content/table.tsx
```diff
@@ -0,0 +1,76 @@
...
+        break;
+
+      default:
+        rows.push(child)
+        break;
+    }
+  })
+
+  return (
+    <table className="govuk-table lbh-table">
```
tcmorris, line 24: Nice on creating a reusable table. Note there is also a variant with a caption included:
https://design-system.hackney.gov.uk/components/table#table-with-head-and-caption

components/announcement.tsx
nathandavies91: At the moment I'm using it to wrap the table, we might need to move away from the page announcement component in the future.

tcmorris (approved):

PR #25, opened 2021-05-20
HRT-31: feature/saving_data

rashthedude:
This PR is part of the `HRT-31` ticket for saving User data to the BE. It currently only saves predetermined mocked data. 

_**Outstanding Tasks**_

- Save form data
- Handle error in case one occurs

### DIFF
--- a/domain/application.ts
+++ b/domain/application.ts
@@ -11,3 +11,9 @@ export interface Application {
   status: string;
   createdAt: string;
 }
+
+export interface ApplicationWithError extends Application {
+  statusCode?: number;
+  error?: string;
+  message?: string;
+}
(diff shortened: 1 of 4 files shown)


next.config.js
```diff
@@ -3,4 +3,8 @@ module.exports = {
     webpack5: true,
   },
   reactStrictMode: true,
+  env: {
+    ENDPOINT_API: process.env.ENDPOINT_API,
+    AWS_KEY: process.env.AWS_KEY
```
tcmorris, line 8: We don't want to expose these values to the front end, as they are sensitive.

domain/application.ts
```diff
@@ -11,3 +11,9 @@ export interface Application {
   status: string;
   createdAt: string;
 }
+
+export interface ApplicationWithError extends Application {
```
tcmorris, line 15: Is this an axios thing? we should get a response or an exception from the API. I wouldn't expect to extend the application here.

tcmorris:
Updated in #26

PR #26, opened 2021-05-21
HRT-31: feature/save-application-data

tcmorris:
- Adds an application to the API once all data is collected (via final submit)
- Uses API route within next.js to act as a proxy for the POST request
- Uses the data that is within the redux store
- Adds some extra fields to the forms
- Adds todos for when we can update an application with the API (currently can only create)

### DIFF
--- a/data/forms/address-details.json
+++ b/data/forms/address-details.json
@@ -1,5 +1,5 @@
 {
-  "heading": "Accomodation details",
+  "heading": "Accommodation details",
   "steps": [
     {
       "fields": [
--- a/data/forms/address-details.json
+++ b/data/forms/address-details.json
@@ -1,0 +1,16 @@
+{
+  "heading": "Accomodation details",
+  "steps": [
+    {
+      "fields": [
+        {
+          "label": "Postcode",
+          "name": "postCode",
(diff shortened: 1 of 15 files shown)


data/forms/address-details.json
```diff
@@ -0,0 +1,16 @@
+{
+  "heading": "Accomodation details",
```
rashthedude, line 2: `Accommodation`

lib/gateways/applications-api.ts
```diff
@@ -34,6 +34,24 @@ export const getApplication = async (id: string): Promise<Application | null> =>
...
+    }
+    const { data } = await axios.post(
+      `${process.env.ENDPOINT_API}/applications`, application,
+      {
+        headers: headers
+      }
+    );
+    return data;
+  } catch (err) {
+    return null;
```
rashthedude, line 51: Do we need care for the potential error or is this just a placeholder for now?

tcmorris: Error handling in general will need updating, so I think this is ok for now.

rashthedude (approved):

PR #27, opened 2021-05-25
Feature/saving more data

rashthedude:
Update(`patch`) on submission and additional residents addition.


lib/gateways/internal-api.ts
```diff
@@ -43,4 +46,42 @@ export const createApplication = async (applicants: Resident[]): Promise<any> =>
...
 
-// TODO: update application
+/**
+ * Update application
+ * @param {Resident[]} applicants The applicant information
+ * @returns {any} The updated application
+ */
+export const updateApplication = async (applicants: Resident[]): Promise<any> => {
+  try {
+    const personal = applicants[0].formData["personal-details"]
```
tcmorris, line 56: This code could be extracted out as it will likely be the same for create/update.

lib/gateways/internal-api.ts
```diff
@@ -43,4 +46,42 @@ export const createApplication = async (applicants: Resident[]): Promise<any> =>
...
+    }
+    const application: CreateApplicationRequest = {
+      status: "Pending",
+      applicant,
+      otherMembers: applicants.length > 1
+        ? applicants.slice(1).map((resident) => resident?.formData["personal-details"])
+        : []
+    }
+
+    const res = await fetch("/api/applications",
```
tcmorris, line 77: Endpoint will require the application id. see swagger docs.

mocks/routes/applications.js
```diff
@@ -75,16 +75,20 @@ const updateApplication = {
...
-        body: applications[0],
+        body: applications[4],
       },
     },
     {
       id: "simulate",
       response: (req, res) => {
         const applicationId = req.params.id;
-        const application = applications.find((applicationData) => applicationData.id === applicationId);
+        const application = applications.find((applicationData) => {
```
tcmorris, line 85: Not sure why this part has changed?

pages/apply/[resident]/[[...step]].tsx
```diff
@@ -52,8 +53,14 @@ const ApplicationStep = (): JSX.Element => {
     }
   }
 
-  const onCompletion = () => {
-    // TODO: call API with patch
+  const onCompletion = async () => {
+    try {
+      // Below line currently returns a 404 as the endpoints arent ready yet
+      await updateApplication([currentResident])
```
Comment (tcmorris) [inline pages/apply/[resident]/[[...step]].tsx:59]: This is only the main applicant, I think we'd want to pass in all applicants similar to create scenario.

mocks/routes/applications.js
rashthedude: `application` would return undefined for some odd reason. The change fixed it for me.

lib/gateways/internal-api.ts
rashthedude: Not entirely sure I understand what you mean.

tcmorris: Currently this code is repeated between create/update methods, we could move to a helper or similar.

tcmorris (approved):

lib/gateways/internal-api.ts
rashthedude: Good shout. Created a helper function now

PR #37, opened 2021-06-15
Determining household size WIP

rashthedude:


components/form/birthdayinput.tsx
```diff
@@ -0,0 +1,90 @@
...
+}: any): JSX.Element {
+  let id = name;
+
+  if (index !== undefined) {
+    id += `.${index}`;
+  }
+
+  return (
+    <div style={{display: "inline-block", "padding": "0 20px 0 0"}}>
+      <label className="govuk-label govuk-date-input__label" htmlFor="dob-day">
```
tcmorris, line 23: `htmlFor={id}`

components/form/birthdayinput.tsx
```diff
@@ -0,0 +1,90 @@
+import FormGroup from './form-group';
+import Hint from './hint';
+import Label from './label';
+import { Field, FieldInputProps, FieldMetaProps } from 'formik';
+import ErrorMessage from './error-message';
+
+
+export function BirthdayInputFields({
```
tcmorris, line 8: Would probably use DateInput rather than Birthday, which is for a specific use case.

components/form/dropdown.tsx
```diff
@@ -0,0 +1,39 @@
...
+        meta: FieldMetaProps<string>;
+      }) => (
+        <FormGroup error={!!meta.touched && !!meta.error}>
+          {label && <Label content={label} strong={true} />}
+          {hint && <Hint content={hint} />}
+          {meta.touched && meta.error && <ErrorMessage message={meta.error} />}
+
+          <div className="govuk-form-group lbh-form-group">
+            <Field as="select" name={name} className="govuk-select lbh-select">
+              <option value="parent">Parent</option>
```
tcmorris, line 29: Looks like these should be configurable.

components/form/birthdayinput.tsx
```diff
@@ -0,0 +1,90 @@
...
+      }) => (
+        <FormGroup error={!!meta.touched && !!meta.error}>
+          {label && <Label content={label} strong={true} />}
+          {hint && <Hint content={hint} />}
+          {meta.touched && meta.error && <ErrorMessage message={meta.error} />}
+
+          <div className="govuk-form-group lbh-form-group">
+            <fieldset className="govuk-fieldset" role="group" aria-describedby="dob-hint">
+              <span id="dob-hint" className="govuk-hint lbh-hint">
+                For example, 31 3 1980
```
tcmorris, line 63: Duplicated hint, I suspect this is where it should be. Also fieldset in design system has a value:
https://design-system.hackney.gov.uk/components/date-input

components/form/birthdayinput.tsx
```diff
@@ -0,0 +1,90 @@
...
+    <Field name={name}>
+      {({
+        field,
+        meta,
+      }: {
+        field: FieldInputProps<string>;
+        meta: FieldMetaProps<string>;
+      }) => (
+        <FormGroup error={!!meta.touched && !!meta.error}>
+          {label && <Label content={label} strong={true} />}
```
tcmorris, line 56: This label can be removed as each sub input has a label.

tcmorris:
This feels like a good example of where a component based form builder would reduce the repetitive nature of building out lots of json with subtle variations. Overall concept seems ok to me though.

components/form/dropdown.tsx
rashthedude: Yes definitely, served as a placeholder until all the options were made available to me. Am going to make it dynamic

PR #42, opened 2021-07-13
Major refactor

angry-dan:
**This PR introduces the following changes:**

1. Remove `whenEligible` HoC because we need a more refined method of controlling the user flow through eligibility
2. Rework the Redux store
    1. Move AWS Cognito elements into their own slice
    2. Prefer the term "Applicant" over "resident"
3. General refactor. tsc should be showing clean.
4. Simplify the data model that we're using to eliminate the duplication of types - especially residents vs applicants. I'm preferring the types from our Swagger definition.
5. Add some helper logic around UK phone numbers
6. Introduce a convention around the storage of arbitrary key/value data that get's attached to an application:
    1. Where possible we'll use the Person, Address and ContactInformation records, but when data doesn't fit into those containers everything else shall be stored using Questions.
    2. Each Question will have an ID in the form `"[stepID]/[fieldName]"`. Where stepID is a reference to the step in the form process
    3. The value for each Question will be stored as a JSON encoded string to enable the API to declare a single data type for this structure.
7. Use Thunks for async actions.
8. Automatically save changes to the Applicant model via a custom Redux middleware.
9. Applications in the housing register API are linked with users in AWS Cognito via a custom property on the AWS Cognito record which references their application ID. This means that each user intentionally can only have a single application in housing register API. It also means we don't have to search in Dynamo for an application by the Cognito user's ID which keeps data clean.

```
+----------------------+                   +------------------------+
|User                  |                   |Application             |
|(AWS Cognito)         |                   |(Housing register API)  |
+----------------------+                   +------------------------+
|sub (PK)              |         +----[∞]->|id                      |
|name                  |         |         |status                  |
|email                 |         |         |mainApplicant           |
|custom:application_id +-[1]-----+         |otherMembers            |
+----------------------+                   +------------------------+
```

**There are also the following tasks outstanding in this PR:**

- [ ] Recreate the address lookup code, and find a better pattern for decorating the forms with functionality than to inject logic into the ApplicationForms and Form components
- [ ] Refactor code like this: `activeStep === 'income-savings'` out of the generic <Form> component by creating a more sensible escape hatch (e.g. a lambda passed in or something)
- [ ] Clean up logic for converting between FormikValues and Applicant's. Centralise the logic that understands "firstName" means Person.firstName whereas "intentionallyHomeless" is Questions
- [ ] ideally we'd only serialise Question values into JSON strings at the API level, not at the store level. That's open for debate though.
- [x] avoid multiple re-renders of the page at bootstrap time whilst we load the cognitoUser and Application record. Speak with Chris on this one as there could be a UI implication. This should fix any remaining issues with initialValues not being applied to forms when you first load them.
- [ ] Complete the redux flows for other members of the household.
- [ ] Recreate the logic in the checkEligible function.
- [x] Complete the redux flows for AWS Cognito - logging in, out, registration etc.
- [ ] Flow for creating a new application
- [ ] @tcmorris - flow for [set password], [verify account], [login] (hold password in memory? allow logging in with just a code?)

**Followup actions:**

- [ ] We have a duplicate of the term "step" - the pages of the form are referred to as steps, but each step can also have a list of steps (although all of our forms currently only have one so far)
- [ ] Have the housing register API be responsible for generating uids for the various records we make.
- [ ] Implement security into the nextJS API layer (Tony's been working on this)
- [ ] Remove all the debug console.logs
- [ ] Some kind of error handling. This is going to be especially important given that save events are dispatched separately from the form flow (we can change that if it makes sense)
- [ ] The eligibility checker is open to buggy states if you set your answers and then go back and change them. My recommendation would be to convert this into a multi-step process so that you only see one question at a time.
- [ ] End to end test
- [ ] convert /Users/danj/Sites/lbh-housing-register/lib/utils/constants.ts (git) to an enum
- [ ] Remove all instances of the term "resident" not in copy.
- [ ] Deploy our frontend app to AWS Amplify.
- [ ] Deploy our AWS Cognito config through code
- [ ] add github actions for PR review process - run tests, eslint and tsc. disable any's
- [ ] Run a full eslint and remove any references to `any` types.


tony-gledhill (commented):
Overview

During the apply process, on the overview step, there is a GUID that is displayed at the top of the page 

http://localhost:3000/apply/overview

tony-gledhill:
There are undefined people being added to the apply process
Household

tcmorris:
Have got it to the point where a new user can sign up and start a new application, with their details initially linked. The `custom:application_id` is mapped to the cognito user after agreeing to terms.
image

I've skipped 'How many people are in this application?' as the designs for this have changed and we allow more people to be added after an application is started at the moment.

The next part I'll look at is getting the sections to load properly, currently they all go to personal details.

PR #43, opened 2021-07-15
Your Situation standalone screen

rashthedude:


pages/apply/[resident]/your-situation.tsx
```diff
@@ -0,0 +1,262 @@
...
+      //console.log('Exit for last question');
+      return true;
+    }
+
+    //console.log('Do not exit form');
+    return false;
+  };
+
+  return (
+    <>
```
Comment (angry-dan) [inline pages/apply/[resident]/your-situation.tsx:213]: Ideally we'd use a <Form> here from components/form

pages/apply/[resident]/your-situation.tsx
```diff
@@ -0,0 +1,262 @@
...
+    //console.log('step Name', stepName);
+
+    const extractedValue: string = formData[stepName];
+    //console.log('extractedValue', extractedValue);
+
+    if (exitForm(extractedValue, stepName)) {
+      router.push(returnHref);
+    }
+
+    const nextQuestion = questionData(extractedValue);
```
Comment (angry-dan) [inline pages/apply/[resident]/your-situation.tsx:180]: I don't think we should confuse the answer that the user gave with the next quesiton that they need to be directed to.

I'd prefer that the `nextQuestion` was a different value stored in the JSON data and that the answer to the original question was simply "yes" or "no" or "true" or "false" or whatever.

For example, instead of:
```
          "options": [
            {
              "label": "Yes",
              "value": "11"
            },
```

what about:
```
          "options": [
            {
              "label": "Yes",
              "value": "yes",
              "nextStep": 11
            },
```

or maybe even better:

```
          "options": [
            {
              "label": "Yes",
              "value": "yes",
              "nextStep": "court-order"
            },
```

PR #51, opened 2021-07-23
HRT-117 - Updated registered overview page

tony-gledhill:
Added background color to information table.
Added missing links
Added SVG icon to button link

### DIFF
--- a/pages/apply/index.tsx
+++ b/pages/apply/index.tsx
@@ -38,7 +38,7 @@ export default function ApplicationHomePage(): JSX.Element {
         housing offer is made.
       </Paragraph>
 
-      <Announcement variant="success">
+      <Announcement variant="info">
         <Table>
           <TableHeading>Bedrooms required</TableHeading>
           <TableHeading>Housing Register</TableHeading>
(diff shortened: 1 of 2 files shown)


tcmorris (approved):

pages/apply/index.tsx
```diff
@@ -38,7 +38,7 @@ export default function ApplicationHomePage(): JSX.Element {
         housing offer is made.
       </Paragraph>
 
-      <Announcement variant="info">
+      <Announcement variant="success">
```
tcmorris, line 41: The design looks to be a faded yellow, you can override with some custom styles if not available in design system.

pages/apply/index.tsx
```diff
@@ -76,8 +76,34 @@ export default function ApplicationHomePage(): JSX.Element {
...
+          >
+            <path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z" />
+          </svg>
+        }
+      >
+        Start now
+      </ButtonLink>
+
+      <Paragraph>
+        <Link href="#">Send me a link to apply later </Link>
```
tcmorris, line 105: Not sure we want to include these links if they don't do anything / go anywhere.

pages/apply/index.tsx
tony-gledhill: I've confirmed with Chris that the background color needs to be blue as there were two versions of the design

pages/apply/index.tsx
tony-gledhill: Chris has said this was added as part of a concept to see if users wanted to have a link sent to them to resume the process, but this is no longer needed. I've had this removed from the design and the code

PR #52, opened 2021-07-26
Feature/hrt 114 personal details

tcmorris:
Adds in missing fields from personal details / start screen. They are pretty much the same, but with a slight difference in wording. Future enhancement would be adding to a shared component.

- title
- gender
- date of birth
- national insurance

Note that title is currently using a predefined list and that 'prefer to self describe' is missing from gender.


data/forms/sign-up-details.json
```diff
@@ -17,12 +44,55 @@
...
         },
         {
-          "label": "Mobile number",
-          "name": "phoneNumber",
+          "as": "dateinput",
+          "label": "Date of birth",
+          "name": "dateOfBirth",
+          "options": [
+            {
+              "label": "Day",
```
angry-dan, line 52: These options should no longer be required with the updated dateInput component

angry-dan (approved):

tony-gledhill (approved):

PR #54, opened 2021-07-26
Feature/HRT-121 address history

angry-dan:


rashthedude (approved):

tony-gledhill (approved):

tcmorris (commented):
Seemed to work ok, however noticed couple of things that could do with updating.
- after submitting address history, it doesn't load values even though complete
- previous address could be more explicit when entering details

PR #58, opened 2021-07-28
Added inset text to medical needs page

tony-gledhill:
Updated page heading to text outlined in the design
Added inset text to yes question to outline a separate medical form needs to be completed if there are medical needs

### DIFF
--- a/data/forms/medical-needs.json
+++ b/data/forms/medical-needs.json
@@ -1,5 +1,5 @@
 {
-  "heading": "Medical Needs",
+  "heading": "Do you have any significant medical needs that affect your housing requirements?",
   "copy": "This will limit the amount of properties you may be suitable for, and can increase wait times. We will discuss your needs and conduct medical checks as part of your application.",
   "steps": [
     {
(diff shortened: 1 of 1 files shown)


data/forms/medical-needs.json
```diff
@@ -21,6 +21,21 @@
           "validation": {
             "required": true
           }
+        },
+        {
+          "name": "proof-of-medical-needs",
+          "as": "insettext",
```
tcmorris, line 27: In the designs this looks to be using announcement styles, but this should be ok.

data/forms/medical-needs.json
```diff
@@ -1,5 +1,5 @@
 {
-  "heading": "Medical Needs",
+  "heading": "Do you have any significant medical needs that affect your housing requirements?",
```
tcmorris, line 2: In the designs, there is a variation for this wording - do you, does <x>. I think it's ok like this however.

tcmorris (approved):

data/forms/medical-needs.json
tony-gledhill: All of the other announcements in the site use the same inset text component. There wasn't a way in the design system to apply the green styling

data/forms/medical-needs.json
tony-gledhill: Is there a way for the JSON data to be contextual to the point in the data capture process?

PR #61, opened 2021-07-29
Added self describe text box

tony-gledhill:


data/forms/person-details.json
```diff
@@ -66,8 +66,17 @@
...
             {
-              "label": "Prefer not to say",
-              "value": ""
+              "label": "Prefer to self-describe",
+              "value": "self"
+            }
+          ]
+        },
+        {
+          "name": "self-describe",
```
tcmorris, line 75: This won't save back to the `gender` property on the person. I think we need a custom field for this, similar to how we've done date input.

tcmorris (changes requested):
We need to ensure the data is saved back to the person, rather than have it as a separate question.

data/forms/person-details.json
```diff
@@ -53,9 +53,9 @@
           }
         },
         {
-          "as": "radios",
+          "as": "radioconditional",
           "label": "Gender",
-          "name": "gender",
+          "name": "gender-value",
```
tcmorris, line 58: This needs to match the property name in `person` object

tcmorris (approved):

PR #63, opened 2021-08-03
Feature/user summary

rashthedude:
User Summary page.


tcmorris (changes requested):
This needs some work. I'd like to move away from having one big page that has individual sections listed in the same file, these should be components. Also, we can't rely on data like this `data[6]['answer']`, we should at the very least be using the property names rather than the position in an array. It's quite likely the question order will be updated whereas we do have some control over the property names we use in the code.

rashthedude:
> This needs some work. I'd like to move away from having one big page that has individual sections listed in the same file, these should be components. Also, we can't rely on data like this `data[6]['answer']`, we should at the very least be using the property names rather than the position in an array. It's quite likely the question order will be updated whereas we do have some control over the property names we use in the code.

Fully agree. That thinking was part of the refactor I had in mind.

tcmorris (approved):

PR #66, opened 2021-08-04
What to expect screen

rashthedude:
Notes: 

- Values are currently hardcoded for this iteration.

### DIFF
--- a/pages/apply/expect.tsx
+++ b/pages/apply/expect.tsx
@@ -18,7 +18,7 @@ const WhatToExpect = (): JSX.Element => {
   };
 
   return (
-    <Layout pageName="Confirmation">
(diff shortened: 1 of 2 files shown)


tcmorris (approved):

pages/apply/expect.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const router = useRouter();
+  const dispatch = useAppDispatch();
+
+  const onCancel = () => {
+    dispatch(signOut());
+    router.push('/');
+  };
+
+  return (
+    <Layout pageName="Confirmation">
```
tcmorris, line 21: Page name is used for `<title>` so should be updated based on current page.

PR #73, opened 2021-08-10
Bedroom logic

rashthedude:
Logic has been derived from @angry-dan algorithm.


lib/utils/bedroomCalculator.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    [5, 'female', 'daughter'],
+    [12, 'female', 'daughter'],
+  ],
+  [
+    [30, 'male', null],
+    [25, 'female', 'partner'],
+  ],
+];
+
+describe.only('BedroomCalculator', () => {
```
angry-dan, line 35: remember to remove the "only" as it prevents all other tests from running.

lib/utils/bedroomCalculator.test.ts
```diff
@@ -0,0 +1,51 @@
...
+  test('[[5, `male`], [5, `male`]] should be awarded 1 bedroom', () => {
+    expect(CALCULATE_BEDROOMS(people[0])).toBe(1);
```
angry-dan, line 37: I think you should inline those `people[n]` references. Using numbered indexes like that just makes the code harder to follow.

lib/utils/bedroomCalculator.ts
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  console.log({ over16, under10, over10 });
+
+  const total = over16 + under10 + over10.reduce((acc, [g, c]) => acc + c, 0);
+
+  return total;
+}
+
+export function CALCULATE_BEDROOMS(people: Array<any>) {
```
angry-dan, line 47: This function really was intended for handling data from a spreadsheet. calculateBedrooms above is useful in the app but I'd delete this one - it's job is just to convert data from the spreadsheet array format into something more usable for processing.

lib/utils/bedroomCalculator.ts
```diff
@@ -0,0 +1,60 @@
...
+      // ...and at least one of the under 10s has this gender then they can share a room.
+      if (matchedUnder10s.length) {
+        // we can eliminate this over 10 from the count - they'll be added in the roundup of under 10s.
+        return [gender, (count - 1) / 2];
+      }
+    }
+    return [gender, count / 2];
+  });
+
+  console.log({ over16, under10, over10 });
```
angry-dan, line 40: debug

pages/apply/expect.tsx
```diff
@@ -7,27 +7,90 @@ import Layout from '../../components/layout/resident-layout';
...
+    let birthDate = new Date(date);
+    let age = today.getFullYear() - birthDate.getFullYear();
+    let m = today.getMonth() - birthDate.getMonth();
+    if (m < 0 || (m === 0 && today.getDate() < birthDate.getDate())) {
+      age--;
+    }
+    return age;
+  };
+
+  const getGender = (gender: string) => {
```
tcmorris, line 37: Should already be a util that does this. See `getGenderName`

pages/apply/expect.tsx
```diff
@@ -7,27 +7,90 @@ import Layout from '../../components/layout/resident-layout';
...
 
+  const [CalculatedBedrooms, setCalculatedBedrooms] = useState(0);
+  const [waitingTime, setWaitingTime] = useState(0);
+
   const onCancel = () => {
     dispatch(signOut());
     router.push('/');
   };
 
+  const getAge = (date: string) => {
```
tcmorris, line 26: Util already exists for this.

lib/utils/bedroomCalculator.ts
```diff
@@ -0,0 +1,60 @@
+function calculateBedrooms(...people: Array<any>) {
```
angry-dan, line 1: We need to type the people array properly. `people` is an array, but it's an array of tuples where the first index is an age and the second is a gender. The correct type to use is `[number, string][]`

lib/utils/bedroomCalculator.test.ts
```diff
@@ -0,0 +1,51 @@
+import { CALCULATE_BEDROOMS } from './bedroomCalculator';
+
+const people = [
```
tcmorris, line 3: Would be useful to add some more test cases here.

lib/utils/bedroomCalculator.ts
```diff
@@ -0,0 +1,60 @@
...
+  // one bedroom each...unless it's a couple
+  let relationshipType = people[0].pop();
```
angry-dan, line 3: Try to avoid using `let`. Also, `pop` is a mutation which should also be avoided.

lib/utils/bedroomCalculator.ts
```diff
@@ -0,0 +1,60 @@
...
+  let over16 = people.filter(([age, _]) => {
+    return age >= 16;
+  }).length;
+
+  // 0.5 bedroom each.
+  const under10 = people.filter(([age]) => age < 10).length / 2;
+
+  if (relationshipType === 'partner') {
+    over16 = over16 - 1;
```
angry-dan, line 12: same here - try to define a new `const` rather than modifying a variable defined with `let`.

pages/apply/expect.tsx
rashthedude: Cheers.

pages/apply/expect.tsx
```diff
@@ -7,27 +7,90 @@ import Layout from '../../components/layout/resident-layout';
...
 import { signOut } from '../../lib/store/cognitoUser';
+import { Applicant } from '../../domain/HousingApi';
+import { CALCULATE_BEDROOMS } from '../../lib/utils/bedroomCalculator';
+import { useState, useEffect } from 'react';
 
 const WhatToExpect = (): JSX.Element => {
   const router = useRouter();
   const dispatch = useAppDispatch();
 
+  const [CalculatedBedrooms, setCalculatedBedrooms] = useState(0);
```
angry-dan, line 18: `CalculatedBedrooms` should have  lower case `c`

pages/apply/expect.tsx
```diff
@@ -7,27 +7,90 @@ import Layout from '../../components/layout/resident-layout';
...
+  const getGender = (gender: string) => {
+    switch (gender) {
+      case 'M':
+        return 'male';
+      case 'F':
+        return 'female';
+    }
+  };
+
+  const getWaitingTime = (bedroom: number) => {
```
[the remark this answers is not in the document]
rashthedude: Sure.

lib/utils/bedroomCalculator.test.ts
rashthedude: Will do.

rashthedude:
I'm refactoring some of the code and will push the changes as soon as.

tcmorris (commented):
Seems to work, added some comments to help tidy up the code.

pages/apply/expect.tsx
```diff
@@ -7,27 +7,90 @@ import Layout from '../../components/layout/resident-layout';
...
+  const getAge = (date: string) => {
+    let today = new Date();
+    let birthDate = new Date(date);
+    let age = today.getFullYear() - birthDate.getFullYear();
+    let m = today.getMonth() - birthDate.getMonth();
+    if (m < 0 || (m === 0 && today.getDate() < birthDate.getDate())) {
+      age--;
+    }
+    return age;
+  };
```
angry-dan, line 35: to calculate an age, wouldn't you just subtract two dates and get the resulting number of years?

pages/apply/expect.tsx
angry-dan, line 37: you don't need to do this. The calculate bedrooms function doesn't process strings like `male` and `female`. It's just looking to see if people have the same or a different gender. You can use anything you like - `M`, `male`, `boy` - it doesn't care as long as you're consistent.

pages/apply/expect.tsx
```diff
@@ -7,27 +7,90 @@ import Layout from '../../components/layout/resident-layout';
...
+  const bedroomArray = applicants.map((individual) => {
+    let age = getAge(individual['person']!['dateOfBirth']!);
+    let gender = getGender(individual['person']!['gender']!);
+    let relationship = individual['person']!['relationshipType']!;
+    return [age, gender, relationship];
+  });
+
+  const bedroom_results = CALCULATE_BEDROOMS(bedroomArray);
+  useEffect(() => {
+    setCalculatedBedrooms(bedroom_results);
```
angry-dan, line 76: You don't want state for this - it's a calculated value so you have two options: 
- calculate it every render
- memoise it with useMemo

Personally I'd just calculate it every render. Unless there's a performance issue that you've spotted it'll keep the code simpler. Don't use state though - you'll introduce an unnecessary render.

angry-dan (changes requested):

pages/apply/expect.tsx
rashthedude: That's exactly what it's doing but in a more comprehensive way.

tcmorris (approved):

PR #81, opened 2021-08-13
Different task options for other household members over 18

rashthedude:


tony-gledhill (approved):

pages/apply/[resident]/index.tsx
```diff
@@ -43,6 +42,13 @@ const ApplicationStep = (): JSX.Element => {
     currentResident === mainResident
   );
 
+  if (
+    currentResident !== mainResident &&
+    getAgeInYears(currentResident) >= 18
```
Comment (tcmorris) [inline pages/apply/[resident]/index.tsx:47]: Think it would be better to pass this is as extra param to `getApplicationSectionsForResident` and do any required changes there so that it's all in one place.

lib/utils/money.ts
```diff
@@ -0,0 +1,25 @@
...
+  heading: string;
+  sections: Sections[];
+}
+
+interface Sections {
+  heading: string;
+  id: FormID;
+}
+
+export const Income: IncomeInterface = {
```
tcmorris, line 13: Shouldn't need to redefine this, we can add to `getApplicationSectionsForResident`

PR #89, opened 2021-08-20
Updated medical needs page

tony-gledhill:
Updated the medical needs page to include a question reveal underneath the title of the page.


rashthedude:
@tony-gledhill there seems to be a problem with the deployment. Could you look into Heroku if you have access to the logs?

tcmorris (approved):

PR #98, opened 2021-08-25
Displaying User errors

rashthedude:


components/errors/user.tsx
```diff
@@ -0,0 +1,21 @@
...
+      className="govuk-error-summary optional-extra-class lbh-error-summary"
+      aria-labelledby="error-summary-title"
+      role="alert"
+      tabIndex={-1}
+      data-module="govuk-error-summary"
+    >
+      <h2 className="govuk-error-summary__title" id="error-summary-title">
+        {children}
+      </h2>
+    </div>
```
tcmorris, line 19: Will likely need to check with design, but I think if we want to use this then we should ensure that it provides the necessary markup from [design system](https://design-system.hackney.gov.uk/components/error-summary). e.g.
```
<h2 className="govuk-error-summary__title" id="error-summary-title">{title}</h2>
<div class="govuk-error-summary__body">
    ...
</div>
```

tcmorris: Reference in another Hackney repo:
https://github.com/LBHackney-IT/lbh-social-care-frontend/blob/main/components/ErrorSummary/ErrorSummary.tsx

pages/apply/start.tsx
```diff
@@ -13,57 +14,65 @@ import {
...
+        })
+      );
 
-    dispatch(createApplication(store.getState().application));
+      dispatch(createApplication(store.getState().application));
 
-    router.push('/apply/verify');
+      router.push('/apply/verify');
+    } catch (error) {
+      setUserError(error.message);
```
tcmorris, line 68: I think we should declare some generic error messages, rather than relying on anything that comes back via an external service.

components/errors/response.tsx
```diff
@@ -0,0 +1,32 @@
+interface ResponseCodes {
+  [key: string]: string;
+}
+
+const ErrorResponseCodes: ResponseCodes = {
+  UsernameExistsException: 'An account with the given email already exists.',
```
tcmorris, line 6: Think it would be better just to have a generic error rather than trying to match on possible exceptions.

rashthedude: That's in place now.

PR #110, opened 2021-09-02
Added main applicant under 18 age check

tony-gledhill:


pages/apply/not-eligible.tsx
```diff
@@ -17,7 +17,7 @@ export default function NotEligible(): JSX.Element {
     (store) => store.application.mainApplicant
   );
   const [isEligible, reasons] = useMemo(
-    () => (mainApplicant && checkEligible(mainApplicant)) ?? [],
+    () => (mainApplicant && checkEligible(mainApplicant, true)) ?? [],
```
tcmorris, line 20: The other pages are checking each member of the application, does this need updating to match?

tcmorris (approved):
PR looks good. Side note, I'm not sure we actually need to check eligibility for other members involved in the application so this can be simplified further.

PR #116, opened 2021-09-06
Feature/hrt 183 personal details

tony-gledhill:


pages/apply/[resident]/personal-details.tsx
```diff
@@ -1,28 +1,58 @@
...
+import * as Yup from 'yup';
+import { Applicant } from '../../../domain/HousingApi';
+import {
+  applicantEqualToOrOlderThanAge,
+  getAgeInYearsFromDate,
+} from '../../../lib//utils/dateOfBirth';
+
+type State = 'under-sixteen' | 'over-sixteen';
+
+export function formatDate(date: Date) {
```
Comment (tcmorris) [inline pages/apply/[resident]/personal-details.tsx:30]: This can be a util.

pages/apply/[resident]/personal-details.tsx
```diff
@@ -71,25 +110,182 @@ const ApplicationStep = (): JSX.Element => {
...
+    switch (state) {
+      case 'under-sixteen':
+        return schema.pick(['title', 'firstName', 'surname', 'dateOfBirth']);
+      case 'over-sixteen':
+        return schema.pick([
+          'title',
+          'firstName',
+          'surname',
+          'dateOfBirth',
+          'nationalInsuranceNumber',
```
Comment (tcmorris) [inline pages/apply/[resident]/personal-details.tsx:222]: We should also need the main applicant scenario where phone / email is required?

tcmorris (commented):
We'll also need this for 'add-person' when creating the household members. Should be able to create a reusable component, which is the form itself.

lib/utils/form.ts
```diff
@@ -52,7 +52,7 @@ export function checkEligible(
   };
 
   if (isMainApplicant) {
-    if (!isOver18(applicant)) {
+    if (!applicantEqualToOrOlderThanAge(applicant, 18)) {
```
tcmorris, line 55: I think this would be better kept as a descriptive function, which has the value `18` set in one place rather than in multiple files.

PR #118, opened 2021-09-07
Email sent confirmation added to App Completion screen

rashthedude:


components/panel.tsx
```diff
@@ -1,13 +1,22 @@
 interface PanelProps {
   heading: string;
   message: string;
+  email: string | undefined;
```
tcmorris, line 4: I think we should update message to be {children} and add the text where the declaration is rather than adding an extra dependency here, which makes it single use case and less reusable.

pages/apply/confirmation.tsx
```diff
@@ -32,6 +35,7 @@ const ApplicationConfirmation = (): JSX.Element => {
       <Panel
         heading="Application complete"
         message={`Your reference number: ${application.reference?.toUpperCase()}`}
+        email={residentEmail}
```
tcmorris, line 38: Set the text here, rather than within the panel.

components/panel.tsx
rashthedude: That's done now

pages/apply/confirmation.tsx
rashthedude: That's done now

PR #121, opened 2021-09-08
Updated current accommodation to include extra questions

tony-gledhill:


components/application/application-forms.tsx
```diff
@@ -39,6 +39,7 @@ export default function ApplicationForms({
   const activeStepId =
     formSteps.find((step) => step === activeStep) ?? formSteps[0];
   const formData = getFormData(activeStepId);
+  console.log('formData', formData);
```
tcmorris, line 42: `console.log` to remove.

tcmorris (commented):
In testing, the values didn't appear to be saving to the redux store so they weren't loading back after submission. (have found a fix)
The actual flow for the forms seemed to work ok.

components/application/application-forms.tsx
tony-gledhill: My bad. Good spot

tony-gledhill:
The values of the first set of questions are being saved under current-accommodation but the others are being saved under current-accommodation-host-details and current-accommodation-landlord-details as the questions are being saved separate files. If we were able to have multiple conditions on the conditionalDisplay field, we could bring the data back into one file

PR #135, opened 2021-09-16
Scroll to top between screen steps

rashthedude:


components/form/form.tsx
```diff
@@ -39,11 +40,13 @@ export default function Form({
 
   const next = () => {
     // TODO: Scroll to top + set focus to first field
```
tcmorris, line 42: Remove TODOs.

PR #142, opened 2021-09-22
HRT-248 current accomodation unsuitable

gdbarnes:


components/form/input.tsx
```diff
@@ -54,6 +54,7 @@ export default function Input({
               placeholder={placeholder}
               type={type}
               {...field}
+              maxLength={100000}
```
tcmorris, line 57: Looks like this should be `10000`?

tcmorris (approved):

components/form/input.tsx
gdbarnes: Yes.. you're right. It was meant to be 10,000.

Will change it to 500 though as it seems like a sensible upper limit on text inputs. 10,000 was just to save the DB.

PR #152, opened 2021-09-27
Feature/hrt 211 mark as sensative

tony-gledhill:
This allows an admin user to set an application as being sensitive. 

There is currently a temporary label to display the initial state of the application. This is temporary as we have not implemented the admin permission system.


tcmorris (commented):
Looks ok to me, can you link the relevant API changes also.

PR #156, opened 2021-09-28
HRT-113: feature/login flow

tcmorris:
Rework the login flow...

**Create verify code**
- enter details and start new application
- create a code linked to application
- send email via API

**Confirm verify code**
- enter code and call API
- get access token with claim on application id
- stored in a cookie

**App init**
- load JWT from cookie and decode with secret
- load application with matching id into store

**Sign out**
- clear store
- clear cookie
- causes redirect back to home


lib/store/application.ts
```diff
@@ -7,12 +7,25 @@ import {
...
 import mainApplicant from './mainApplicant';
 import otherMembers from './otherMembers';
 import { NotifyRequest, NotifyResponse } from '../../domain/govukNotify';
 
+export const importApplication = createAsyncThunk(
+  'applicant/import',
+  async (application: Application) => {
+    return application as Application;
+  }
+);
```
angry-dan, line 27: This doesn't seem quite right. A thunk would usually perform some kind of async action. This function appears to just return what it's given.

lib/store/application.ts
```diff
@@ -32,6 +45,35 @@ export const createApplication = createAsyncThunk(
   }
 );
 
+export const createVerifyCode = createAsyncThunk(
```
angry-dan, line 48: I might not yet understand this properly - but I imagine the verification code logic should live in a separate store. This file is really meant to hold all of the logic around the application itself.

lib/store/application.ts
```diff
@@ -32,6 +45,35 @@ export const createApplication = createAsyncThunk(
...
+  'auth/create',
+  async (application: Application) => {
+    const request: CreateAuthRequest = {
+      email: application.mainApplicant?.contactInformation?.emailAddress ?? '',
```
angry-dan, line 52: So the `?`s here are a bit of a code smell. I made that MainApplicant type to help with this. The MainApplicant type has assurances over the email address being present I think. Although actually there's a good chance you want a string argument here for the email address rather than needing the whole application passed in.

angry-dan: Also I'd recommend trying to be consistent in those naming conventions and maybe stick to `emailAddress` everywhere?

lib/utils/users.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (err instanceof jsonwebtoken.JsonWebTokenError) {
+      return;
+    }
+
+    throw err;
+  }
+}
+
+export const signOut = (): void => {
+  // TODO: clear cookie
```
angry-dan, line 30: don't forget to actually todo this.

pages/apply/verify.tsx
```diff
@@ -30,16 +32,19 @@ const ApplicationVerifyPage = (): JSX.Element => {
 
   const confirmSignUp = async (values: FormData) => {
     try {
-      await Auth.confirmSignUp(emailAddress, values.code);
+      //await Auth.confirmSignUp(emailAddress, values.code);
```
angry-dan, line 35: We should probably tidy up any references to AWS Cognito, deleting code like this and any unused functions.

lib/utils/users.ts
```diff
@@ -0,0 +1,32 @@
+import cookie from 'cookie';
+import jsonwebtoken from 'jsonwebtoken';
+import { HackneyResident } from '../../domain/HackneyResident';
+
```
angry-dan, line 4: I feel like we need a clearer separation than this `utils` structure is allowing.
The `getUser` function appears to only execute server side - it's called from the getServerSideProps function and it won't work in the browser.
On the other hand - `signOut` is running in the browser. I think the problem is that these functions both have to be executed in their respective environments - `getUser` accesses an env var and `signOut` uses the browser's `window` object.

angry-dan: Also - and maybe the build processes have improved these days - but in the past this kind of thing would cause all the `imports` to appear in the client side build and make your JS files big.

pages/api/auth/[id]/generate.tsx
```diff
@@ -0,0 +1,31 @@
+import { StatusCodes } from 'http-status-codes';
+import type { NextApiHandler, NextApiRequest, NextApiResponse } from 'next';
+import { createVerifyCode } from '../../../../lib/gateways/applications-api';
+
```
Comment (angry-dan) [inline pages/api/auth/[id]/generate.tsx:4]: This file name feels a touch inconsistent - since it's executing an API endpoint called `createVerifyCode`

pages/api/auth/[id]/verify.tsx
```diff
@@ -0,0 +1,48 @@
+import { StatusCodes } from 'http-status-codes';
+import type { NextApiHandler, NextApiRequest, NextApiResponse } from 'next';
+import { confirmVerifyCode } from '../../../../lib/gateways/applications-api';
+import cookie from 'cookie';
+
```
Comment (angry-dan) [inline pages/api/auth/[id]/verify.tsx:5]: This file name is also a bit non-obvious. Maybe go long and call it `confirmVerifyCode` or `confirm-verify-code`.

pages/api/auth/[id]/verify.tsx
```diff
@@ -0,0 +1,48 @@
...
+              maxAge: (7 * 24 * 60 * 60),
+              domain: '.hackney.gov.uk',
+              path: '/',
+            }
+          );
+
+          res.setHeader('Set-Cookie', jwtCookie);
+        }
+
+        res.status(StatusCodes.OK).json(data);
```
Comment (angry-dan) [inline pages/api/auth/[id]/verify.tsx:32]: You're returning all of data here - is that intentional? Are there any secrets or things that shouldn't be sent to the browser? Also thinking ahead in case the JWT library made an API change in the future.

pages/apply/start.tsx
```diff
@@ -64,7 +68,11 @@ const ApplicationStartPage = (): JSX.Element => {
         })
       );
 
-      dispatch(createApplication(store.getState().application));
+      dispatch(createApplication(store.getState().application)).then(
+        (action: any) => {
```
angry-dan, line 72: The any here feels unnecessary.

pages/apply/overview.tsx
```diff
@@ -10,22 +10,38 @@ import SummaryList, {
...
+
+const ApplicationPersonsOverview = ({data}: PageProps): JSX.Element => {
   const router = useRouter();
   const dispatch = useAppDispatch();
 
+  useEffect(() => {
+    dispatch(
+      importApplication(data)
+    );
+  }, []);
```
angry-dan, line 43: So I see the approach that you're going for here - preload the application with the page load and then put it straight into the store at the first render.

Redux does actually let you pre-fill the store when you create it, it's just that we're initing the store in the App rather than on each page. This data that you're loading has wider context than just this page though - so perhaps you need a getServerSideProps for the app and then you can take care of the whole thing there.

tcmorris: Ok thanks, have switched the code to load the application in the app init similar to how it worked before.

lib/store/application.ts
tcmorris: Have moved.

lib/utils/users.ts
tcmorris: Yep understood, have updated the functions to all be server side.

PR #158, opened 2021-09-29
Feature/hrt 203 medical form

tony-gledhill:
Link medical form to application. A text input allowing me to paste in the Google Form URL so that I can link it to the application, and a date input to confirm when the form was received.

An update to the API project is also needed for this to be completed

### DIFF
--- a/components/applications/admin-form.tsx
+++ b/components/applications/admin-form.tsx
@@ -34,10 +34,12 @@ export default function AdminForm({
 
   const schema = buildValidationSchema(step.fields);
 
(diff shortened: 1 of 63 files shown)


tcmorris (commented):
Every input field should have a label for accessibility, not sure why we've added `displayLabel`?

tony-gledhill:
I re-used the same form controls ion the front end as the admin site. Everything was fine, but the labels were appearing in the wrong place.

tcmorris (approved):

PR #163, opened 2021-09-29
First pass at a CSV export tool.

angry-dan:
A few things concerning me about this approach.
- We need a system that can scale up to at least 10k rows to be sustainable. AWS API gateway is going to enforce a 30s timeout on requests and currently I'm storing the entire CSV output in memory. In other words, we're pushing our luck - soon enough we'll either hit a timeout or memory issue...
- ... and rightly so. This sort of thing should be handled in a batch. At the very least it should be a queued job that get's notified on completion. The user shouldn't be connected to the process that's building the file. Once it's done it should get stored on S3 for a while.
- we're currently pulling data in chunks of 20 pages. And we're pulling way more data than we need. We'll need bigger pages of less data.
- There are a lot of unused or not understood columns, including some extra data points we need to be capturing.
A few things concerning me in general:
- package lock appears to be out of date
- tsc isn't passing at the moment, a few test files are out of date


lib/utils/auth.ts
```diff
@@ -49,3 +51,16 @@ export const getRedirect = (
     return '/access-denied';
   }
 };
+
+export function getAuth(
```
tcmorris, line 55: This looks similar to code above, might need a refactor with other admin pages.

pages/api/applications/generate-report.tsx
```diff
@@ -0,0 +1,174 @@
...
+Address3                          MainApplicant
+Address4                          MainApplicant
+Address5                          MainApplicant
+Postcode                          MainApplicant
+HomeTelephone                     MainApplicant
+WorkTelephone                     [we're not collecting]
+Email                             ''
+NINumber                          MainApplicant
+Sex                               "M" or "F"
+RegistrationDate                  ? Date application made or approved?
```
tcmorris, line 116: I think this is the submission date currently.

pages/api/applications/generate-report.tsx
```diff
@@ -0,0 +1,174 @@
...
+EffectiveDate                     ? Effective Band Date.
```
tcmorris, line 117: This will be the 'application date', which can be edited by an officer and could be backdated.

pages/api/applications/generate-report.tsx
```diff
@@ -0,0 +1,174 @@
...
+ApplicantType                     ? Could this be band? Sample data included URG and RES as values. Urgent?
+MinimumBedSize                    Calculated bedroom count minus one min zero.
+MaximumBedSize                    Calculated bedroom count
+DateOfBirth                       MainApplicant
+OlderPersonsAssessement           ''
+MobilityAssessment                ''
+AdditionalBandingInfo             ''
+MedicalRequirements               ?
+Offered                           ? status = 4 then "Y" , 8 then "N"
+EthnicOrigin                      ? Encoded AOO and TCY
```
tcmorris, line 127: We ask the question for this, but don't have the mappings to the codes.

tcmorris:
Appreciate the comments and notes from above @angry-dan I think this is still valuable to add in as it is, so will merge. As you mention the API should return better data which is linked to reporting purposes. I don't think the intention is to export all of the data at once, so in reality this is likely to be a time period.

PR #173, opened 2021-10-01
HRT-260 - Redirect users to the not-eligible page if disqualified

gdbarnes:
Also small copy and styling change on not eligible page

### DIFF
--- a/pages/apply/overview.tsx
+++ b/pages/apply/overview.tsx
@@ -1,3 +1,4 @@
+import { useMemo } from 'react';
 import { ButtonLink } from '../../components/button';
 import { HeadingOne } from '../../components/content/headings';
 import Paragraph from '../../components/content/paragraph';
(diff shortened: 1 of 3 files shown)


tcmorris (approved):
I think this is ok, note there will be a temp flash before the page redirects.

pages/apply/overview.tsx
```diff
@@ -34,6 +37,19 @@ const ApplicationPersonsOverview = (): JSX.Element => {
       .flat()
   );
 
+  const mainApplicant = useAppSelector(
+    (store) => store.application.mainApplicant
+  );
+
+  const [isEligible] = useMemo(
```
tcmorris, line 44: Check status of the application instead, requires updating the application via API on disqualification.

tcmorris (approved):

PR #176, opened 2021-10-01
Added admin permissions for sensitive applications

tony-gledhill:


pages/applications/view/[id]/index.tsx
```diff
@@ -59,94 +62,104 @@ export default function ApplicationPage({
   return (
     <UserContext.Provider value={{ user }}>
       <Layout pageName="View application">
-        {data.sensitiveData && (
+        {data.sensitiveData &&
+        !user.hasAdminPermissions &&
+        !user.hasManagerPermissions &&
+        !user.hasOfficerPermissions ? (
```
Comment (tcmorris) [inline pages/applications/view/[id]/index.tsx:68]: If they are an officer they might not be able to view the application, but we can check if they are assigned.
e.g. `(user.hasOfficerPermissions && data.assignedTo == user.email)`

tcmorris (approved):

PR #184, opened 2021-10-02
Feature/view register

tcmorris:
- Adds extra page to view all applications.
- Updates search to use the new page by default

Note, this depends on an [API update](https://github.com/LBHackney-IT/lbh-housing-api/pull/26).

### DIFF
--- a/components/admin/application-table.tsx
+++ b/components/admin/application-table.tsx
@@ -9,7 +9,7 @@ import { formatDate } from '../../lib/utils/dateOfBirth';
 
 interface TableProps {
   caption?: string;
-  applications: PaginatedApplicationListResponse;
+  applications: PaginatedApplicationListResponse | null;
   currentPage: number;
   pageUrl: string;
   parameters: URLSearchParams;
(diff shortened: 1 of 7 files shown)


components/admin/application-table.tsx
```diff
@@ -24,7 +24,7 @@ export default function ApplicationTable({
 }: TableProps): JSX.Element {
   return (
     <>
-      {applications.results.length > 0 ? (
+      {applications?.results.length > 0 ? (
```
angry-dan, line 27: Typescript should probably have noticed this could be null/undefined. If it didn't then it suggests the type of `applications` is wrong somewhere. 
In `TableProps`consider making `applications` optional with a `?`
Also in the function params consider using `applications = []` then you could avoid the `?` here which is a bit neater - it's nicer if you know that values are going to exist.

pages/applications/unassigned.tsx
```diff
@@ -136,7 +119,7 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
 
   const applications =
     reference === '' && status === ''
-      ? await getApplications(page)
+      ? await getApplications(page, 'unassigned')
```
angry-dan, line 122: That magic string should probably be defined somewhere? An easy, good enough answer might be:
```
function getApplications(page, query: string | "unassigned")
```

components/admin/application-table.tsx
tcmorris: Declared that it could be null and added extra check.

PR #186, opened 2021-10-04
Feature/bidding number

tcmorris:
- Add extra validation on bidding number
- Add option to generate bidding number

### DIFF
--- a/components/admin/actions.tsx
+++ b/components/admin/actions.tsx
@@ -2,7 +2,7 @@ import * as Yup from 'yup';
 import { Formik, Form, FormikValues } from 'formik';
 import { Application } from '../../domain/HousingApi';
 import Button from '../button';
-import DateInput from '../form/dateinput';
+import DateInput, { INVALID_DATE } from '../form/dateinput';
 import Select from '../form/select';
 import Radios from '../form/radios';
 import Input from '../form/input';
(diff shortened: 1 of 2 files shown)


components/admin/actions.tsx
```diff
@@ -209,10 +213,11 @@ export default function Actions({ data }: PageProps): JSX.Element {
         band: values.band,
         reason: values.reason,
         biddingNumber: values.biddingNumber,
+        generateBiddingNumber: values.biddingNumberType === 'generate',
       },
     };
     updateApplication(request);
-    router.reload();
+    _.delay(() => router.reload(), 500);
```
angry-dan, line 220: You don't need lodash for this...
```
setTimeout(() => router.reload(), 500)
```

Although this kind of thing is usually a code smell. What's taking 500ms that you need to wait for?

components/admin/actions.tsx
```diff
@@ -182,11 +183,14 @@ export default function Actions({ data }: PageProps): JSX.Element {
     reason: Yup.string()
       .label('Reason')
       .oneOf(reasonOptions.map(({ value }) => value)),
-    applicationDate: Yup.string(),
-    informationReceived: Yup.string(),
+    applicationDate: Yup.string().notOneOf([INVALID_DATE], 'Invalid date'),
+    informationReceived: Yup.string().notOneOf([INVALID_DATE], 'Invalid date'),
```
angry-dan, line 187: I think we need to centralise the logic with our date fields and the validation. Also, what's `Invalid date` if it's not `INVALID_DATE`?

components/admin/actions.tsx
tcmorris: Yep agree, it was added due to the page reloading the data before it's persisted.

components/admin/actions.tsx
tcmorris: I took this from `address-history.tsx`, could definitely do with tidying up validation in general.

PR #187, opened 2021-10-05
Feature/add gtm

tcmorris:
Add google tag manager.


pages/_app.tsx
```diff
@@ -13,6 +14,7 @@ function App({ Component, pageProps }: AppProps): ReactElement {
 
   const dispatch = useAppDispatch();
   useEffect(() => {
+    TagManager.initialize({ gtmId: process.env.NEXT_PUBLIC_GTM_ID! });
```
angry-dan, line 17: Something seems odd about this approach. I've never used the library before, but you usually have to embed GTM in the page head. If it's going in a useEffect then I'm guessing it's not loading GTM until React is bootstrapped and that's going to cause inaccurate page load time values in Google Analytics as well as make adding tags to the page load event difficult.

serverless.yml
```diff
@@ -43,6 +43,7 @@ functions:
       LOOKUP_API_TOKEN: ${ssm:/housing-register/${self:provider.stage}/lookup-api-key}
       NEXT_PUBLIC_ENV: ${opt:stage}
       NEXT_PUBLIC_EVIDENCE_STORE: ${ssm:/housing-register/${self:provider.stage}/evidence-store-url}
+      NEXT_PUBLIC_GTM_ID: ${ssm:/housing-register/${self:provider.stage}/gtm-id}
```
angry-dan, line 46: It'd be nice for a future project to understand if it's possible to import all the variables from a .env.sample file to save the need to duplicate them in here. for now, maybe should have a comment at the top of the sample file to say "don't forget to add your variables to serverless.yml too"

pages/_app.tsx
tcmorris: Details here: https://www.npmjs.com/package/react-gtm-module
I actually noticed this was used elsewhere in a Hackney repo, but yeah alternative I guess would be to add to the `_document.tsx`

PR #195, opened 2021-10-07
HRT-228 - evidence requests

tcmorris:
Create evidence request on submission.

Mirrors the changes made in the API: https://github.com/LBHackney-IT/lbh-housing-api/pull/28

### DIFF
--- a/domain/HousingApi.ts
+++ b/domain/HousingApi.ts
@@ -484,3 +484,12 @@ export interface VerifyAuthRequest {
 export interface VerifyAuthResponse {
   accessToken: string;
 }
+
+export interface CreateEvidenceRequest {
+  userRequestedBy?: string;
+  documentTypes: Array<string>;
+}
+export interface EvidenceRequestResponse {
+  id: string;
+  createdAt: string;
+}
--- a/domain/HousingApi.ts
+++ b/domain/HousingApi.ts
@@ -487,7 +487,7 @@ export interface VerifyAuthResponse {
 
 export interface CreateEvidenceRequest {
   userRequestedBy?: string;
-  documentTypes: Array<string>;
+  documentTypes: string[];
 }
 export interface EvidenceRequestResponse {
   id: string;
--- a/lib/utils/evidence.ts
+++ b/lib/utils/evidence.ts
@@ -1,0 +1,25 @@
+import { Applicant } from '../../domain/HousingApi';
+import { getQuestionValue } from '../store/applicant';
+import { FormID } from './form-data';
+
+export const getRequiredDocumentsForApplication = (
+  applicant: Applicant
+): Array<string> => {
+  let documentTypes = [
+    'proof-of-id',
+    'proof-of-address',
+    'proof-of-income',
+    'proof-of-savings',
+  ];
+
+  const employment = getQuestionValue(
+    applicant.questions,
+    FormID.EMPLOYMENT,
+    'employment'
+  );
+  if (employment === 'retired') {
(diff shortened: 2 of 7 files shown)


domain/HousingApi.ts
```diff
@@ -484,3 +484,12 @@ export interface VerifyAuthRequest {
 export interface VerifyAuthResponse {
   accessToken: string;
 }
+
+export interface CreateEvidenceRequest {
+  userRequestedBy?: string;
+  documentTypes: Array<string>;
```
angry-dan, line 490: convention with arrays is usually:
```
string[]
```
It's the same as what you have, just syntactic sugar - thought I'd mention because it appears a few times in the PR but not urgent to change it.

lib/utils/evidence.ts
```diff
@@ -0,0 +1,25 @@
+import { Applicant } from '../../domain/HousingApi';
+import { getQuestionValue } from '../store/applicant';
+import { FormID } from './form-data';
+
+export const getRequiredDocumentsForApplication = (
+  applicant: Applicant
+): Array<string> => {
+  let documentTypes = [
```
angry-dan, line 8: fun fact: you could still use `const` here. 

Because `documentTypes` itself isn't changing when you `push` to it below. Just the data held inside of it. Typescript has a handy `as const` that you can add if you want to make objects that can't be mutated but it's rarely necessary.
If you wanted to avoid the mutation all together then this would also work:

```
return  [
    'proof-of-id',
    'proof-of-address',
    'proof-of-income',
    'proof-of-savings',
   ...(employment === 'retired' ? ['proof-of-pension'] : [])
];

```

I'll let you decide which is the neater of the two.

pages/api/applications/[id]/evidence.tsx
```diff
@@ -0,0 +1,31 @@
...
+import type { NextApiHandler, NextApiRequest, NextApiResponse } from 'next';
+import { createEvidenceRequest } from '../../../../lib/gateways/applications-api';
+
+const endpoint: NextApiHandler = async (
+  req: NextApiRequest,
+  res: NextApiResponse
+) => {
+  switch (req.method) {
+    case 'POST':
+      try {
```
Comment (angry-dan) [inline pages/api/applications/[id]/evidence.tsx:11]: There doesn't seem to be any access checking going on here or elsewhere? Does that mean that anyone could set up evidence requests for an application?

lib/utils/evidence.ts
angry-dan: those strings should also be from some kind of enum in the domain model I reckon

pages/apply/submit/declaration.tsx
```diff
@@ -40,6 +41,7 @@ const Declaration = (): JSX.Element => {
       }
 
       dispatch(completeApplication(application));
+      dispatch(createEvidenceRequest(application));
```
angry-dan, line 44: I'm still not exactly sure with Redux thunks, but I have a feeling that maybe both of these posts are meant to go in the same thunk? So you could dispatch one action and then fire both of the requests from the one thunk handler (or have two handlers for the same action but that doesn't seem such a good idea).
This is probably absolutely fine, I can't see any benefit to be hand from changing it around at this stage.

angry-dan (changes requested):

pages/api/applications/[id]/evidence.tsx
Comment (angry-dan) [reply pages/api/applications/[id]/evidence.tsx:11]: read the cookie
check that the JWT application id  matches the [id] in the url

PR #219, opened 2021-11-03
HRT-306 - Login flow

gdbarnes:
Should allow user to save and exit, and then log back in to an incomplete application using their email address after verification.

### DIFF
--- a/pages/apply/sign-in.tsx
+++ b/pages/apply/sign-in.tsx
@@ -8,15 +8,12 @@ import { FormData } from '../../lib/types/form';
 import { FormID, getFormData } from '../../lib/utils/form-data';
 import ErrorSummary from '../../components/errors/error-summary';
 import Paragraph from '../../components/content/paragraph';
-// import { createVerifyCode } from '../../lib/gateways/applications-api';
 import { createVerifyCode } from '../../lib/store/auth';
-import { CreateAuthRequest } from '../../domain/HousingApi';
 
 const ApplicationSignInPage = (): JSX.Element => {
   const dispatch = useAppDispatch();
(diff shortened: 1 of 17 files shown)


pages/apply/sign-in.tsx
```diff
@@ -22,16 +23,20 @@ const ApplicationSignInPage = (): JSX.Element => {
   }, [isLoggedIn]);
 
   const onSubmit = async (values: FormData) => {
-    // TODO create a verify code based on email provided and send to verify page
-    //dispatch(createVerifyCode(values.emailAddress));
-    //router.push('/apply/verify');
+    dispatch(createVerifyCode(values.emailAddress));
```
tcmorris, line 26: Add try/catch with `setUserError`

pages/apply/start.tsx
```diff
@@ -41,19 +34,12 @@ const ApplicationStartPage = (): JSX.Element => {
             nationalInsuranceNumber: values.nationalInsuranceNumber,
```
tcmorris, line 34: Think `updateBeforeFirstSave` might need to be updated to `updateApplicant`

tcmorris (commented):
Changes look ok to me, have tested briefly but haven't covered both scenarios.

pages/apply/sign-in.tsx
gdbarnes: try/catch added

pages/apply/start.tsx
gdbarnes: I've put this in but haven't been able to run a true new application. We might need to ensure that the main applicant's id is available at this stage.

PR #221, opened 2021-11-04
bug:hrt333 - bedroom calculation now retrieved from api

jasattilia:


domain/HousingApi.ts
```diff
@@ -483,3 +489,6 @@ export interface EvidenceRequestResponse {
   id: string;
   createdAt: string;
 }
+export interface SimpleTypeResponse<T> {
```
tcmorris, line 492: Doesn't look like this is used?

tcmorris (approved):

domain/HousingApi.ts
jasattilia: it was when we're returning an calculated bedroom number from the api. I've left it in there because the Bedroom calculator endpoint returns this..

PR #235, opened 2021-11-24
HRT-303 - resident facing overview

gdbarnes:


styles/generic/_overrides.scss
```diff
@@ -73,3 +77,40 @@ body {
 .lbh-link--announcement:visited {
   color: #0b0c0c;
 }
+
+.applicant-summary {
```
tcmorris, line 81: Add new styles to component folder.

tcmorris (approved):

styles/generic/_overrides.scss
gdbarnes: Fair point. Moved to partial.

PR #241, opened 2021-11-30
HRT-210 - add case manually

gdbarnes:


tcmorris (commented):
Looks good and seems to work for the most part. Have pushed a couple of tweaks to show/hide the edit button and assessment options.

Main issue that I noticed was that the address history wasn't showing up in the summary section after being saved.

PR #279, opened 2022-03-15
Feature/HRT-432 don't disqualify medical

angry-dan:


lib/utils/form.ts
```diff
@@ -65,24 +63,20 @@ export function checkEligible(application: Application): [boolean, string[]] {
...
-  //******** *//
 
   if (!isOver18(mainApplicant)) {
-    setInvalid(disqualificationReasonOptions.under18YearsOld);
+    setInvalid('under18YearsOld');
   }
 
-  for (const [form, values] of Object.entries(FormID)) {
-    const eligibilityCriteria = getEligibilityCriteria(values);
+  for (const [, formID] of Object.entries(FormID)) {
```
gdbarnes, line 74: I realise this is some legacy code not part of this PR, but if we only need the value from the enum maybe we use Object.values instead of Object.entries:
```suggestion
  for (const formID of Object.values(FormID)) {
```

gdbarnes (approved):

PR #293, opened 2022-04-05
HRT-431 - Application complete page redirects incorrectly

gdbarnes:


lib/store/application.ts
```diff
@@ -56,10 +56,9 @@ export const disqualifyApplication = createAsyncThunk(
 export const completeApplication = createAsyncThunk(
   'application/complete',
   async (application: Application) => {
-    const res = await fetch(`/api/applications/${application.id}/complete`, {
+    await fetch(`/api/applications/${application.id}/complete`, {
       method: 'PATCH',
     });
```
angry-dan, line 61: I'm still a little uncomfortable with this solution.

The real reason for this bug is that the API endpoint which we're hitting is returning a 500 server error.
When we were doing `return (await res.json()) as Application;` the problem was that the JSON coming back wasn't in fact an Application, but a description of the error, and that in turn was causing our store to see an error message as the actual application which has bad consequences.

What we really need to do is have all our calls to `fetch` also check for the `.ok` prop (https://developer.mozilla.org/en-US/docs/Web/API/Response/ok) before returning an application. If it's not `ok` then it should throw an error. That way the promise will reject, and in turn Redux will see a rejected state on the thunk.

I'd suggest that whatever we do to this one endpoint should also be applied to all the other places we're using thunk/fetch together too.

PR #295, opened 2022-04-05
HRT-434 - add case manually submitted date

gdbarnes:
Adds complete application to internal api
Generates submitted date when addin a manual application

### DIFF
--- a/pages/applications/add-case.tsx
+++ b/pages/applications/add-case.tsx
@@ -5,7 +5,11 @@ import { getRedirect, getSession } from '../../lib/utils/googleAuth';
 import { HackneyGoogleUser } from '../../domain/HackneyGoogleUser';
 import { FormikValues } from 'formik';
 import { Application } from '../../domain/HousingApi';
-import { createApplication } from '../../lib/gateways/internal-api';
+import {
+  createApplication,
+  completeApplication,
+  updateApplication,
+} from '../../lib/gateways/internal-api';
 import { generateQuestionArray, Address } from '../../lib/utils/adminHelpers';
 import { ApplicationStatus } from '../../lib/types/application-status';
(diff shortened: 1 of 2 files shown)


pages/applications/add-case.tsx
```diff
@@ -55,12 +57,22 @@ export default function AddCasePage({ user }: PageProps): JSX.Element {
...
+    createManualApplication(request);
+  };
+
+  const createManualApplication = async (request: Application) => {
+    try {
+      const newApplication = await createApplication(request);
+      const completedApplication = await completeApplication(newApplication);
+      const setToManualDraft = await updateApplication({
+        ...completedApplication,
+        status: ApplicationStatus.MANUAL_DRAFT,
```
angry-dan, line 69: This definitely feels like one of those situations where @anefiox could help you out to streamline this flow. But I guess this would work too.

pages/applications/add-case.tsx
```diff
@@ -55,12 +57,22 @@ export default function AddCasePage({ user }: PageProps): JSX.Element {
...
       });
-    });
+
+      router.push(`/applications/view/${setToManualDraft.id}`);
+    } catch (error) {
+      console.error(error);
```
angry-dan, line 74: What's the benefit of catching this error like this?

angry-dan (approved):

pages/applications/add-case.tsx
gdbarnes: Will remove this.

pages/applications/add-case.tsx
gdbarnes: Agreed. Maybe a future ticket so we don't have to call 3 different endpoints.

PR #299, opened 2022-04-07
HRT-442 - Adds delete link for household members

gdbarnes:


components/admin/other-members.tsx
```diff
@@ -48,6 +60,27 @@ export default function OtherMembers({
...
+                </button>
+
+                <Dialog
+                  isOpen={showDeleteWarningDialog}
+                  title="Confirm delete"
+                  onConfirmation={() => handleDelete(applicantToDelete)}
+                  onCancel={() => setShowDeleteWarningDialog(false)}
+                >
+                  <Paragraph>
+                    Are you sure you want to delete{' '}
```
angry-dan, line 78: Can I check in on that wording "delete"? I'm no UXer, but I think maybe "remove" might be better?

angry-dan (approved):

components/admin/other-members.tsx
gdbarnes: No worries, I'll change it to remove!

PR #301, opened 2022-04-08
HRT-440 - update edit statuses

gdbarnes:
Adds in new logic for permissions and statuses allowed to edit applications. Removes unused imports.

### DIFF
--- a/pages/applications/view/[id]/index.tsx
+++ b/pages/applications/view/[id]/index.tsx
@@ -39,6 +39,7 @@ import {
   getRedirect,
   getSession,
   HackneyGoogleUserWithPermissions,
+  hasAnyPermissions,
 } from '../../../../lib/utils/googleAuth';
 import { getPersonName } from '../../../../lib/utils/person';
 import Custom404 from '../../../404';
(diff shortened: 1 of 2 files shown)


pages/applications/view/[id]/index.tsx
```diff
@@ -56,15 +57,32 @@ export default function ApplicationPage({
...
+      return true;
+    }
+    if (
+      data.status === ApplicationStatus.AWAITING_REASSESSMENT &&
+      assignedToCurrentUser
+    ) {
+      return true;
+    }
+    return false;
+  };
```
Comment (angry-dan) [inline pages/applications/view/[id]/index.tsx:84]: This is great, but I do think you might need to replicate some similar logic at the API level to ensure that these permissions are properly enforceable

Comment (gdbarnes) [reply pages/applications/view/[id]/index.tsx:87]: Maybe putting this and other admin permissions logic at the API level should be in a future ticket?

The reason being that similar tests already exist such as `canViewSensitiveApplication` in `googleAuth.ts` which use this convention (checking the properties on `user`).

angry-dan (approved):
Happy to take your judgement on this

PR #302, opened 2022-04-11
HRT-438 - changing status

gdbarnes:


components/admin/actions.tsx
```diff
@@ -267,6 +51,15 @@ export default function Actions({ data }: PageProps): JSX.Element {
         'aboveZero',
         'Bedroom need should be a number greater than 0',
         (value, testContext) => {
+          console.log(value, testContext);
```
angry-dan, line 54: debug!

components/admin/actions.tsx
```diff
@@ -267,6 +51,15 @@ export default function Actions({ data }: PageProps): JSX.Element {
...
+
+          if (
+            testContext.parent.status !== ApplicationStatus.ACTIVE &&
+            testContext.parent.status !== ApplicationStatus.ACTIVE_UNDER_APPEAL
+          ) {
+            return true;
```
angry-dan, line 60: This validation condition doesn't seem to match the error message: "Bedroom need should be a number greater than 0"

components/admin/actions.tsx
```diff
@@ -341,94 +134,138 @@ export default function Actions({ data }: PageProps): JSX.Element {
...
     }
 
-    updateApplication(request);
-    setTimeout(() => router.reload(), 500);
+    try {
+      updateApplication(request).then(() => {
+        router.reload();
+      });
+    } catch (error) {
+      console.log('error', error);
```
angry-dan, line 142: what's the benefit of catching this error? For context I'm setting up Sentry monitoring at the moment which will notice when errors get thrown and record them centrally.

pages/applications/view/[id]/index.tsx
```diff
@@ -56,8 +56,28 @@ export default function ApplicationPage({
   data,
   history,
 }: PageProps): JSX.Element | null {
-  if (!data.id) return <Custom404 />;
   const router = useRouter();
```
Comment (angry-dan) [inline pages/applications/view/[id]/index.tsx:59]: I'm pretty sure that `useRouter` is a state based hook, which means that every time that `router.query.tab` changes, this component will be re-rending anyway. If I'm right, then I think you could skip the state all together and just do:

```
  const tab = router.query.tab ?? 'overview';

  const handleTabChange = (newValue: string) => {
    router.push(`/applications/view/${data.id}?tab=${newValue}`, undefined, {
      shallow: true,
    });
  };
```

And each time you call handleTabChange you should be able to rely on tab updating. In other words, router.push is internally doing a setState so you probably don't need a second one yourself.

angry-dan (approved):
Happy to approve but a couple of observations. Up to you if you need to make changes.

components/admin/actions.tsx
gdbarnes: The idea here is to skip validation (return true) when the selected status doesn't show the bedroom need question. Maybe there's a better way but I'm not sure what it is.

components/admin/actions.tsx
gdbarnes: Will remove

components/admin/actions.tsx
gdbarnes: Will remove

pages/applications/view/[id]/index.tsx
Comment (gdbarnes) [reply pages/applications/view/[id]/index.tsx:59]: Makes sense. Have updated. We could probably use this shallow routing on other places in the admin but maybe for another PR.

PR #386, opened 2024-08-06
Pin serverless to version 3

LBHTKarki:
Pin serverless to version 3 to test whether that resolves some permission issues in the pipeline.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ jobs:
       - *attach_workspace
       - run:
           name: deploy
-          command: npm run build && sudo npm i -g serverless && sls deploy --stage development
+          command: npm run build && sudo npm i -g serverless@^3 && sls deploy --stage development
 
   build-deploy-staging:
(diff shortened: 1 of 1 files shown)


sam-drumm (approved):

.circleci/config.yml
```diff
@@ -52,23 +52,23 @@ jobs:
       - *attach_workspace
       - run:
           name: deploy
-          command: npm run build && sudo npm i -g serverless && sls deploy --stage development
+          command: npm run build && sudo npm i -g serverless@^3 && sls deploy --stage development
```
Duslerke, line 55: Can also do
``` yaml
command: |
   npm run build
   npm i -g serverless@^3
   sls deploy --stage developmnet
```
to avoid those `&&`s

LBHTKarki: Yes, that would be a bit tidier. I'll leave it for now since it doesn't affect the functionality and the PR itself fixed the issue I was looking into.

PR #391, opened 2024-08-07
Linting configuration

sam-drumm:
https://hackney.atlassian.net/browse/TS-1586?atlOrigin=eyJpIjoiOGYzNjA3OTNjYjIwNGY2OWI5YWM0NWE2OTk4YzgwYzEiLCJwIjoiaiJ9

**What**
- Adds linter configuration from Hackney and necessary libraries, 
- Adds a precommit step to check tests, tsc, lint and prettify staged files
- Adds coverage reports for Jest so that CircleCI can show results and Sonar can run analysis
- Adds .nvmrc 

**Why**
As linting rules are not applied to this directory, we will have a lot of errors and warnings to work through if we apply it to the whole code base. 

This PR allows us to have better feedback in the IDE, but only force rules on staged files. 

**How**
- Husky will run a precommit step to lint-fix/ prettify-fix staged files, run tests, run tsc. Commit will fail if not passed.


.eslintrc
```diff
@@ -0,0 +1,64 @@
...
+  },
+  "ignorePatterns": ["mocks/**", "coverage/**"],
+  "rules": {
+    "linebreak-style": ["error", "unix"],
+    "require-atomic-updates": "off",
+    "react/react-in-jsx-scope": "off",
+    // "react/function-component-definition": ["warn"],
+    // "react/jsx-no-useless-fragment": ["warn"],
+    // "jsx-a11y/anchor-is-valid": ["warn"],
+    // "eqeqeq": ["warn", "smart"],
```
Duslerke, line 34: perhaps, remove if commented?

package-lock.json
Duslerke, line 1: What happened here?

Duslerke (approved):

Duslerke:
Approved, but I recommend resolving the package lock diff now

package-lock.json
sam-drumm: Yeah good catch didn't notice that one thanks

sam-drumm: Interesting that I can push commits to this branch after approval and it doesn't go stale....

.circleci/config.yml
```diff
@@ -131,7 +141,7 @@ workflows:
       - sonar-scan:
           context: SonarCloud
           requires:
-            - install-dependencies
+            - tests
```
sam-drumm, line 144: Requires test rather than in parallel because it relies on the coverage reports.

sam-drumm:
I understand this PR would make this.. 

https://github.com/LBHackney-IT/lbh-housing-register/blob/TS-1586-Housing-reg-view-only-Review-linting/.pre-commit-config.yaml

...redundant - but want to review that.

LBHTKarki:
Now that we are using husky I guess we could remove the pre-commit config file `.pre-commit-config.yaml`

LBHTKarki (approved):

PR #394, opened 2024-08-16
Ts 1585 housing reg view only implement new view only group in the app

sam-drumm:
https://hackney.atlassian.net/browse/TS-1585?atlOrigin=eyJpIjoiNmUyMTk4ZjkzNGI3NDllZjkyYjgxZDhmNDc3ZGJhNWIiLCJwIjoiaiJ9

*Note utility function changes outlined in ticket will be addressed in a seperate PR.

**What** 
- Implements a readOnly user type in googleAuth.ts that can be used to show or hide features based on cookie status. This follows a similar pattern to the way other features are shown for different user groups. If a user has readOnly group which isn't superseded by another higher group (any other group) then it will apply the readOnly status.
- Implements an e2e tests to prove this case, 
- As a product of this implementation, where linting errors have been raised components have been refactored to account for this,
- Updates the pipeline to run e2e tests against a build rather than the dev environment,
- Adds a Sentry option to the next.config that ensures sentry is run only in production. 

**Why**
- Service wants a read-only user to search for applications but have no other functionality

**Future**
- Investigate further the best way to mock SeverSide props on a next app in Cypress. Currently it's pointing to the dev env api.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -25,7 +25,6 @@
       "version": "detect"
     }
   },
-  "ignorePatterns": ["mocks/**", "coverage/**", "next.config.js"],
   "rules": {
     "linebreak-style": ["error", "unix"],
     "require-atomic-updates": "off",
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -2,17 +2,6 @@ import { defineConfig } from 'cypress';
 import 'dotenv/config';
 import { sign } from 'jsonwebtoken';
 
-// type Environment = 'localdev' | 'development';
-
-// const environment: Environment =
-//   (process.env.NEXT_PUBLIC_ENV as Environment) || 'development';
-
-// const baseUrlSites: Record<Environment, string> = {
-//   localdev: 'http://localhost:3000',
-//   development: 'https://housing-register-development.hackney.gov.uk/',
-// };
-
-// const baseUrl = baseUrlSites[environment];
 const baseUrl = 'http://localhost:3000';
 
 export default defineConfig({
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -1,5 +1,6 @@
 import { defineConfig } from 'cypress';
 import 'dotenv/config';
+import { sign } from 'jsonwebtoken';
 
 type Environment = 'localdev' | 'development';
 
@@ -17,9 +18,16 @@ export default defineConfig({
   e2e: {
     baseUrl,
     experimentalWebKitSupport: true,
-    setupNodeEvents() // on, config
-    {
-      // implement node event listeners here
+    setupNodeEvents(on, config) {
+      on('task', {
+        generateToken({ user, secret }) {
+          return sign(user, secret);
+        },
+      });
+      config.env = {
+        ...process.env,
+      };
+      return config;
     },
     video: true,
     screenshotOnRunFailure: true,
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -2,17 +2,18 @@ import { defineConfig } from 'cypress';
 import 'dotenv/config';
 import { sign } from 'jsonwebtoken';
 
-type Environment = 'localdev' | 'development';
+// type Environment = 'localdev' | 'development';
 
-const environment: Environment =
-  (process.env.NEXT_PUBLIC_ENV as Environment) || 'development';
+// const environment: Environment =
+//   (process.env.NEXT_PUBLIC_ENV as Environment) || 'development';
 
-const baseUrlSites: Record<Environment, string> = {
-  localdev: 'http://localhost:3000',
-  development: 'https://housing-register-development.hackney.gov.uk/',
-};
+// const baseUrlSites: Record<Environment, string> = {
+//   localdev: 'http://localhost:3000',
+//   development: 'https://housing-register-development.hackney.gov.uk/',
+// };
 
-const baseUrl = baseUrlSites[environment];
+// const baseUrl = baseUrlSites[environment];
+const baseUrl = 'http://localhost:3000';
 
 export default defineConfig({
   e2e: {
--- a/cypress/e2e/searchForAnApplication.cy.ts
+++ b/cypress/e2e/searchForAnApplication.cy.ts
@@ -4,6 +4,7 @@ import ApplicationsPage from '../pages/applications';
 
 describe('Search for an application', () => {
   it('as a read only user I only see the search results', () => {
+    cy.clearAllCookies();
     cy.loginAsUser('readOnly');
     ApplicationsPage.visit();
 
@@ -27,6 +28,7 @@ describe('Search for an application', () => {
     ApplicationsPage.getWorktraySidebar().should('not.exist');
   });
   it('as an officer I can engage with worktray', () => {
+    cy.clearAllCookies();
     cy.loginAsUser('officer');
     ApplicationsPage.visit();
 
--- a/cypress/e2e/searchForAnApplication.cy.ts
+++ b/cypress/e2e/searchForAnApplication.cy.ts
@@ -1,0 +1,52 @@
+import { faker } from '@faker-js/faker/locale/en_GB';
+
+import ApplicationsPage from '../pages/applications';
+
+describe('Search for an application', () => {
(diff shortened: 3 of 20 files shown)


cypress.config.ts
```diff
@@ -1,25 +1,34 @@
 import { defineConfig } from 'cypress';
 import 'dotenv/config';
+import { sign } from 'jsonwebtoken';
 
-type Environment = 'localdev' | 'development';
+// type Environment = 'localdev' | 'development';
 
-const environment: Environment =
-  (process.env.NEXT_PUBLIC_ENV as Environment) || 'development';
+// const environment: Environment =
```
LBHTKarki, line 7: I guess all these comments can be removed

cypress/e2e/searchForAnApplication.cy.ts
```diff
@@ -0,0 +1,54 @@
+import { faker } from '@faker-js/faker/locale/en_GB';
+
+import ApplicationsPage from '../pages/applications';
+
+describe('Search for an application', () => {
+  it('as a read only user I only see the search results', () => {
+    // cy.clearAllCookies();
```
LBHTKarki, line 7: comment can be removed

cypress/e2e/searchForAnApplication.cy.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    // can see the search results
+    ApplicationsPage.getSearchResults().should('be.visible');
+
+    // doesn't see the sidebar
+
+    ApplicationsPage.getWorktraySidebar().should('not.exist');
+  });
+  it('as an officer I can engage with worktray', () => {
+    // cy.clearAllCookies();
```
LBHTKarki, line 31: comment can be removed

cypress/support/commands.ts
```diff
@@ -1,5 +1,11 @@
 import { faker } from '@faker-js/faker/locale/en_GB';
 import { mount } from 'cypress/react';
+// import 'dotenv/config';
```
LBHTKarki, line 3: comment can be removed

cypress/support/helpers.ts
```diff
@@ -1,13 +1,14 @@
 export const screenPresets: Cypress.ViewportPreset[] = [
   // Desktops
-  'macbook-15', // Set viewport to 1440px x 900px
+  'macbook-16', // set viewport to 1536 x	960
+  // 'macbook-15', // Set viewport to 1440px x 900px
```
LBHTKarki, line 4: comments can be removed

lib/utils/googleAuth.ts
```diff
@@ -1,31 +1,63 @@
...
     if (!parsedToken) return;
 
-    var secret = process.env.HACKNEY_JWT_SECRET as string;
-    const user = (
-      process.env.SKIP_VERIFY_TOKEN !== 'true'
-        ? jsonwebtoken.verify(parsedToken, secret)
-        : jsonwebtoken.decode(parsedToken)
-    ) as HackneyGoogleUser | undefined;
+    const secret = process.env.HACKNEY_JWT_SECRET as string;
+    const user = (process.env.SKIP_VERIFY_TOKEN !== 'true'
```
Duslerke, line 58: Hopefully this happens server-side :sweat_smile:

cypress/support/commands.ts
```diff
@@ -55,3 +61,42 @@ Cypress.Commands.add('generateEmptyApplication', () => {
...
+  if (!user) {
+    throw new Error(`No user data found for user type "${userType}"`);
+  }
+  const secret = 'aDummySecret';
+
+  cy.task('generateToken', { user, secret }).then((token) => {
+    const cookieName = 'hackneyToken';
+    cy.getCookies().should('be.empty');
+    cy.setCookie(cookieName, token as string);
+    cy.getCookie(cookieName).should('have.property', 'value', token);
```
Duslerke, line 100: I don't think this assertion can fail unless cypress were to drop support for setCookie :thinking: .
Or is this a case of cypress not supporting setCookie on some browsers, but not others?

cypress/support/helpers.ts
Duslerke, line 1: Will just ask whether this is intentional? _(it could be just some leftover changes from all those test commits)_.

Duslerke (commented):
PR looks good.

Though, the commit messages are not very descriptive & as such not very useful.

I recommend **Squash Merge'ing** this PR for cleaner git history. Also, best time to squash is from now (from feature to dev) ... as the squash merge from development to main will be troublesome afterwards.

.eslintrc
```diff
@@ -25,7 +25,6 @@
       "version": "detect"
     }
   },
-  "ignorePatterns": ["mocks/**", "coverage/**", "next.config.js"],
```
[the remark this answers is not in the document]
sam-drumm: Yeah good point - I actually removed this because we have a .eslintignore and these are not really patterns. So I've made sure coverage and mocks is in that file (now) just to avoid confusion.

package.json
```diff
@@ -20,8 +20,11 @@
     "component:run": "cypress run --component --browser electron"
   },
   "lint-staged": {
-    "*.{js,jsx,ts,tsx}": "eslint --fix --max-warnings 0",
-    "!package-lock.json": "prettier --ignore-unknown --write"
+    "*.{js,jsx,ts,tsx}": [
+      "eslint --fix --max-warnings 0",
+      "prettier --ignore-unknown --write"
+    ],
+    "!(*.js|*.jsx|*.ts|*.tsx|package-lock.json)": "prettier --ignore-unknown --write"
```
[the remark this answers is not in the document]
sam-drumm: lint-fix and then make 'prettier' lint js,jsx,ts,tsx files

If not js,jsx,ts,tsx just prettier lint them

cypress/support/helpers.ts
sam-drumm: The helpers file - yeah for some viewports.

cypress/support/commands.ts
sam-drumm: Haven't given this a lot of thought aside from just checking the cookie set exists in the browser environment? Or am I missing something.

pages/applications/index.tsx
```diff
@@ -46,61 +53,67 @@ export default function ApplicationListPage({
...
+                  isActive={activeNavItem === 'Pending'}
+                >
+                  Pending
+                </HorizontalNavItem>
+              </HorizontalNav>
+              {applications ? (
+                <>
+                  <ApplicationsTable
+                    applications={applications}
+                    showStatus
```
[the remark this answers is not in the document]
sam-drumm: yeah it's truthy when just the prop like that.

package.json
```diff
@@ -5,10 +5,10 @@
   "scripts": {
     "build": "next build",
     "dev": "next dev --hostname=localdev.hackney.gov.uk",
-    "start": "next start -p $PORT",
+    "start": "next start",
     "mocks": "mocks-server",
     "lint": "eslint . --max-warnings 0 --ext .js,.jsx,.ts,.tsx",
-    "lint:fix": "eslint . --fix --max-warnings 0 --ext .js,.jsx,.ts,.tsx",
```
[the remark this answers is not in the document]
sam-drumm: yeah good catch.

sam-drumm:
Excuse the bunch of commits for this - was testing (and failing) with some pipeline issues. 

Will be squshed on merge.

package.json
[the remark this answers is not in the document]
Duslerke: Mind bender :sweat_smile: 
Cool, that makes sense, thanks for clarifying!

cypress/support/commands.ts
Duslerke: It's probably fine. My comment was more so a question of whether there's any circumstance for when line 100 could fail?

As this reads somewhat like:

```cs
var token = "some string"
string myCookie = token;
myCookie.Should().Be(myCookie);
```

In essence, the assertion tests Cypress's in-built internal functionality. So I was wondering whether there are any cases when `cy.setCookie` is expected to fail to set the cookie?

Duslerke (approved):

LBHTKarki (approved):

PR #395, opened 2024-08-20
TS-1585 add test coverage for auth utils

LBHTKarki:
## WHAT
Add test coverage for googleAuth and related functions.

Files covered:
1. lib/utils/googleAuth.spec.ts
2. lib/utils/requestAuth.spec.ts
3. lib/utils/users.spec.ts

## WHY
Currently we don't have any test coverage for the functions involved in authentication. In order to develop the application further we need to add appropriate test coverage for the functions and components we modify. 

This adds baseline test coverage for the functions that are involved in authentication.

## HOW
Please note these tests are not opinionated about the current implementation, but cover the functionality as is. 

`requestAuth.ts` tests are lacking some mock setups due to the module structure which makes it difficult. Since this initial update is done without any refactoring or changes to the implementation this function is tested mainly against real implementatations.

jest and lint configs have been updated to ignore the new `testUtils` folder. Also some linting overrides have been added to test helpers to overcome some import issues that we haven't been able resolve in any other way yet.

### DIFF
--- a/lib/utils/googleAuth.spec.ts
+++ b/lib/utils/googleAuth.spec.ts
@@ -15,6 +15,7 @@ import {
   AUTHORISED_ADMIN_GROUP_TEST,
   AUTHORISED_MANAGER_GROUP_TEST,
   AUTHORISED_OFFICER_GROUP_TEST,
+  AUTHORISED_READONLY_GROUP_TEST,
   UserRole,
   generateHRUserWithPermissions,
   generateSignedTokenByRole,
@@ -22,7 +23,9 @@ import {
 } from '../../testUtils/userHelper';
 import {
   HackneyGoogleUserWithPermissions,
+  Permissions,
   canViewSensitiveApplication,
+  canViewWorktray,
   getAuth,
   getPermissions,
   getRedirect,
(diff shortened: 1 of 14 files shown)


testUtils/userHelper.ts
```diff
@@ -0,0 +1,160 @@
...
+  Admin,
+  ReadOnly,
+}
+
+interface UserClaims {
+  hasAdminPermissions: boolean;
+  hasManagerPermissions: boolean;
+  hasOfficerPermissions: boolean;
+  hasReadOnlyPermissions: boolean;
+}
```
Duslerke, line 29: This looks like a copy of `type Permissions` in the google auth file.

testUtils/userHelper.ts
```diff
@@ -0,0 +1,160 @@
...
+    }
+    default:
+      return {
+        hasAdminPermissions: false,
+        hasManagerPermissions: false,
+        hasOfficerPermissions: false,
+        hasReadOnlyPermissions: false,
+      };
+  }
+};
```
Duslerke, line 92: Could probably reduce its size a lot:

```ts
export const getClaimsByRole = (role?: UserRole): UserClaims => {
  const userPerms = {
        hasAdminPermissions: false,
        hasManagerPermissions: false,
        hasOfficerPermissions: false,
        hasReadOnlyPermissions: false,
      };

   switch (role) {
    case UserRole.ReadOnly: {
      userPerms.hasReadOnlyPermissions = true;
      break;
    }
    case UserRole.Officer: {
      userPerms.hasOfficerPermissions: true;
      break;
    }
    case UserRole.Manager: {
      userPerms.hasManagerPermissions = true;
      break;
    }
    case UserRole.Admin: {
      userPerms.hasAdminPermissions = true;
      break;
    }
    default:
      break; 
  }
  return userPerms;
};
```

lib/utils/googleAuth.spec.ts
```diff
@@ -0,0 +1,483 @@
...
+      ];
+
+      const user = generateJWTTokenTestData(groupClaims, issuedAt);
+
+      const expectedPermissions: Permissions = {
+        hasAdminPermissions: true,
+        hasManagerPermissions: true,
+        hasOfficerPermissions: true,
+        hasReadOnlyPermissions: true,
+      };
```
Duslerke, line 300: Can the user have more than 1 set of permissions?

LBHTKarki: Technicaly they could be in multiple groups, but of course that's not how the permissions are setup in practise. However this particular test is just for checking that the claims are set correctly based on the provided groups.

testUtils/userHelper.ts
LBHTKarki: A lot tidier that way 👍 Updated in  [#0213be2](https://github.com/LBHackney-IT/lbh-housing-register/pull/395/commits/0213be25f6dcb0d99b84e50bf04aa55b15ca655a)

Duslerke (approved):

PR #399, opened 2024-09-02
TS-1571 lock down pages for read only users

LBHTKarki:
## WHAT
Currently read only users can access certain edit and other pages directly. They should not be able to access these pages at all.

## WHY
Read only users are authorized to view only certain pages and data. This update locks down the pages they shouldn't be able to access.

## HOW
1. Update `getRedirect ` function with additional read only authorization check that'll direct users to access denied page if they are not allowed direct access to the page
2. Update the following pages by implementing the read only authorization check
2.1  `applications/edit/{applicationId}/{personId}/edit-household-member`
2.2 `applications/edit/{applicationId}/add-household-member`
2.3 `applications/View-register`
2.4 `applications/unassigned`
2.5 `applications/add-case`
3. Disable eslint rules where fixing the issue would require refactoring or changes to existing variable values

## FUTURE
Address the eslint issues by updating the code where required.

### DIFF
--- a/lib/utils/googleAuth.ts
+++ b/lib/utils/googleAuth.ts
@@ -133,12 +133,16 @@ export const hasReadOnlyPermissionOnly = (
 };
 
 export const getRedirect = (
-  user?: HackneyGoogleUserWithPermissions
+  user?: HackneyGoogleUserWithPermissions,
+  checkForReadOnlyPermissions?: boolean
 ): string | undefined => {
   if (!user) {
     return '/login';
   }
-  if (!hasAnyPermissions(user)) {
+  if (
+    !hasAnyPermissions(user) ||
+    (checkForReadOnlyPermissions && hasReadOnlyPermissionOnly(user))
+  ) {
     return '/access-denied';
   }
 };
--- a/lib/utils/googleAuth.ts
+++ b/lib/utils/googleAuth.ts
@@ -134,14 +134,14 @@ export const hasReadOnlyPermissionOnly = (
 
 export const getRedirect = (
   user?: HackneyGoogleUserWithPermissions,
-  checkForReadOnlyPermissions?: boolean
+  writePermissionsRequired?: boolean
 ): string | undefined => {
   if (!user) {
     return '/login';
   }
   if (
     !hasAnyPermissions(user) ||
-    (checkForReadOnlyPermissions && hasReadOnlyPermissionOnly(user))
+    (writePermissionsRequired && hasReadOnlyPermissionOnly(user))
   ) {
     return '/access-denied';
   }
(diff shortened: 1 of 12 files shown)


lib/utils/googleAuth.ts
```diff
@@ -133,12 +133,16 @@ export const hasReadOnlyPermissionOnly = (
 };
 
 export const getRedirect = (
-  user?: HackneyGoogleUserWithPermissions
+  user?: HackneyGoogleUserWithPermissions,
+  checkForReadOnlyPermissions?: boolean
```
Duslerke, line 137: Don't we always want to ensure that users with read-only permissions can do only what's allowed?

lib/utils/googleAuth.ts
```diff
@@ -133,12 +133,16 @@ export const hasReadOnlyPermissionOnly = (
...
 ): string | undefined => {
   if (!user) {
     return '/login';
   }
-  if (!hasAnyPermissions(user)) {
+  if (
+    !hasAnyPermissions(user) ||
+    (checkForReadOnlyPermissions && hasReadOnlyPermissionOnly(user))
+  ) {
```
Duslerke, line 145: Perhaps, this would be better? Then all the hardcoded `true` arguments and the param can be avoided.
```ts
if (!hasAnyPermissions(user) || hasReadOnlyPermissionOnly(user)) { /*...*/ }
```

lib/utils/googleAuth.ts
Duslerke: Nevermind, I see what the issue is: the `getRedirect` is used by both the pages that the read-only user is allowed on, and the on the pages that they're not allowed on.

So this function needs a way to distinguish between whenever they should be allowed on and when not. Perhaps, renaming this variable would be useful? Like:

```ts
writePermissionsRequired?: boolean
```

lib/utils/googleAuth.ts
Duslerke: nvm, see the line 136 comment

LBHTKarki: Because the `getRedirect` function is shared between different page types at the moment it's sometimes fine for the user to have read only permissions. For example they can view the application by accessing the view application page, but they can't edit the application by accessing the edit application page. Both pages use the same function and flow to check whether redirect is required. With this setup the responsibility for the check is in the calling page, hence the new argument. We are looking at refactoring the whole authorization flow to make things easier to maintain and also to shift the responsibility of the logic into a single module.

lib/utils/googleAuth.ts
LBHTKarki: Yes, I think that's a bit more suitable name for it. Updated in https://github.com/LBHackney-IT/lbh-housing-register/pull/399/commits/794a0841e84e41b379f783a633b71f6a7d501b61

sam-drumm (approved):

Duslerke (approved):

PR #401, opened 2024-09-03
TS-1569 - remove edit main applicant access from read only users

LBHTKarki:
## WHAT
Currently read only users can access the edit main applicant page directly. This access must be removed.

## WHY
We don't want read only users editing main applicant details.

## HOW
Implement updated `getRedirect` function that supports checking for read only permissions.

## FUTURE
Resolve the linting issues

### DIFF
--- a/pages/applications/edit/[id]/[person]/index.tsx
+++ b/pages/applications/edit/[id]/[person]/index.tsx
@@ -1,23 +1,27 @@
 import { useState } from 'react';
+
+import { FormikValues } from 'formik';
 import { GetServerSideProps } from 'next';
 import { useRouter } from 'next/router';
-import { FormikValues } from 'formik';
-import {
-  Application,
-  Address as ApiAddress,
-} from '../../../../../domain/HousingApi';
-import { getApplication } from '../../../../../lib/gateways/applications-api';
-import { getRedirect, getSession } from '../../../../../lib/utils/googleAuth';
-import { updateApplication } from '../../../../../lib/gateways/internal-api';
-import Custom404 from '../../../../404';
+
+import MainApplicantForm from '../../../../../components/admin/MainApplicantForm';
 import { HackneyGoogleUser } from '../../../../../domain/HackneyGoogleUser';
 import {
+  Address as ApiAddress,
+  Application,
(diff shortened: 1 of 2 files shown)


LBHJHeppinstall (approved):

pages/applications/edit/[id]/[person]/index.tsx
```diff
@@ -94,12 +98,12 @@ export default function EditApplicant({ user, data }: PageProps): JSX.Element {
...
     setIsSubmitted(true);
   };
-
+  /* eslint-disable react/jsx-no-useless-fragment */
   return (
     <>
       {data.id ? (
         <MainApplicantForm
-          isEditing={true}
+          isEditing={true} //eslint-disable-line
```
Comment (sam-drumm) [inline pages/applications/edit/[id]/[person]/index.tsx:106]: Out of interest, what error was throwing here?

pages/applications/edit/[id]/[person]/index.tsx
```diff
@@ -94,12 +98,12 @@ export default function EditApplicant({ user, data }: PageProps): JSX.Element {
 
     setIsSubmitted(true);
   };
-
+  /* eslint-disable react/jsx-no-useless-fragment */
   return (
     <>
```
Comment (sam-drumm) [inline pages/applications/edit/[id]/[person]/index.tsx:103]: Really curious how on some of these fragments, if you remove the empty div it has an issue with the data. Can't quite get my head around why expect maybe the linter is reading it wrong.

sam-drumm (approved):

pages/applications/edit/[id]/[person]/index.tsx
Comment (LBHTKarki) [reply pages/applications/edit/[id]/[person]/index.tsx:106]: this is where it changes it from `isEditing={true}` to`isEditing` which feels a bit dangerous without further investigation

pages/applications/edit/[id]/[person]/index.tsx
Comment (LBHTKarki) [reply pages/applications/edit/[id]/[person]/index.tsx:103]: I just disabled that for now, but definitely worth investigating further at later date

pages/applications/edit/[id]/[person]/index.tsx
Comment (sam-drumm) [reply pages/applications/edit/[id]/[person]/index.tsx:106]: Oh that's right - that's the one from yesterday.

PR #402, opened 2024-09-03
Clean up Ts 1561 housing reg view only tests

sam-drumm:
https://hackney.atlassian.net/browse/TS-1561

**What**
e2e Tests failing on the pipeline when merging into development, but not locally. This addresses that issue. 

**Future**
Note the application is not currently mocked - future tickets in the sprit will address that.

### DIFF
--- a/cypress/e2e/viewAnApplication.cy.ts
+++ b/cypress/e2e/viewAnApplication.cy.ts
@@ -1,7 +1,7 @@
 import ApplicationsPage from '../pages/applications';
 import { screenPresets } from '../support/helpers';
 
-describe('Application view page', () => {
+describe('View a resident application', () => {
   screenPresets.forEach((screenPreset) => {
     it(`as a read only group user I cannot edit application details on ${screenPreset}`, () => {
       cy.task('clearNock');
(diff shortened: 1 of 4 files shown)


cypress/e2e/viewAnApplication.cy.ts
```diff
@@ -1,18 +1,17 @@
...
 
-describe('Application view page', () => {
+describe('View a resident application', () => {
   screenPresets.forEach((screenPreset) => {
     it(`as a read only group user I cannot edit application details on ${screenPreset}`, () => {
-      cy.viewport(screenPreset);
-
       cy.task('clearNock');
       cy.clearCookies();
+      cy.viewport(screenPreset);
```
1JW1, line 9: Just curious, the positioning of this has shifted from near the top of the test structure to around the middle, what issue does doing this address?

sam-drumm: I don't really have a very scientific answer on that one. The issue was that the Nock (which is intercept for the server side props) was failing in the PL and just wouldn't hit the page. So really was a guess that it might need to do the task first to keep it happy.

cypress/e2e/viewAnApplication.cy.ts
```diff
@@ -45,17 +44,17 @@ describe('Application view page', () => {
...
       cy.loginAsUser('manager');
 
       ApplicationsPage.visit();
       ApplicationsPage.getSearchInput().should('be.visible');
-      ApplicationsPage.getSearchInput().type('1');
+      ApplicationsPage.getSearchInput().type('John');
```
Duslerke, line 57: Based on the PR description it seems like the PR s tackling some sort of tests flakiness.

This change, however, seems to change the test itself. The before and after do not seem equivalent.
Does this change solve some sort of issue?

sam-drumm: This specific change is because pipeline test is connected to the dev api, but we're building the mocks/fixtures in a following ticket. 

The reason for this change is that the '1' record on dev api is not complete and in parts has bad data. So this is just a temp fix.

Duslerke (approved):

PR #403, opened 2024-09-04
TS-1567 hide money and health sections for read only users

LBHTKarki:
##WHAT
Currently read only users can view applicants health and finance details. This update hides those details.

##WHY
We don't want read only users to see those details. 

##HOW
Stop money and health components from rendering when user has read only permissions. New application and user helpers have been added, however these only cover properties required by this feature. 

##FUTURE
Enhance the objects generated by the helpers to support more tests.


sam-drumm (approved):
LGTM

pages/applications/view/[id]/[person]/index.tsx
```diff
@@ -157,20 +157,32 @@ export default function ApplicationPersonPage({
...
+                      itemName="money"
+                      isActive={activeNavItem === 'money'}
+                      dataTestId="test-applicant-money-section-navigation"
+                    >
+                      Money
+                    </HorizontalNavItem>
+                  ) : (
+                    // eslint-disable-next-line react/jsx-no-useless-fragment
+                    <></>
+                  )}
```
Comment (Duslerke) [inline pages/applications/view/[id]/[person]/index.tsx:172]: This is a valid linting message - it shouldn't be disabled.
The whole expression can be rewritten to avoid useless fragment:

```ts
!hasReadOnlyPermissionOnly(user) && (
   <HorizontalNavItem
      handleSelectNavItem={handleSelectNavItem}
      itemName="money"
      isActive={activeNavItem === 'money'}
      dataTestId="test-applicant-money-section-navigation"
    >
    Money
    </HorizontalNavItem>
)
```

Comment (LBHTKarki) [reply pages/applications/view/[id]/[person]/index.tsx:172]: I don't seem to be able to use that expression wihout wrapping it in to a fragment which results in the same error. What am I missing?

Comment (LBHTKarki) [reply pages/applications/view/[id]/[person]/index.tsx:172]: Sorry should have included the update. This is what I tried:
```
{!hasReadOnlyPermissionOnly(user) && (
  <HorizontalNavItem
    handleSelectNavItem={handleSelectNavItem}
    itemName="money"
    isActive={activeNavItem === 'money'}
    dataTestId="test-applicant-money-section-navigation"
  >
    Money
  </HorizontalNavItem>
)}
```
That results in `Type 'false | Element' is not assignable to type 'Element'.  Type 'boolean' is not assignable to type 'ReactElement<any, any>'.ts(2322)`. That expression is widely used though, so not sure what's wrong with it in this case.

Comment (LBHTKarki) [reply pages/applications/view/[id]/[person]/index.tsx]: If I wrap all these child `HorizontalNavItem` components in a fragment then it works. Perhaps that's the way to go. Like this https://github.com/LBHackney-IT/lbh-housing-register/pull/403/commits/9607ca85e49f57438da87256a76516d7eedcca8d

Comment (Duslerke) [reply pages/applications/view/[id]/[person]/index.tsx]: Ehh.. the problem is typescript complaining that the return type of the expression is variable and depends on how the condition evaluates. IF false, the returns bool, if true the returns JSX.

Seems like the 2 common solutions are:
1. Do the ternary like you did, except return `null` instead of fragment.
2. Wrap the whole expression into fragment to force an unconditional JSX return type.

Comment (Duslerke) [reply pages/applications/view/[id]/[person]/index.tsx]: > If I wrap all these child `HorizontalNavItem` components in a fragment then it works. Perhaps that's the way to go. Like this [9607ca8](https://github.com/LBHackney-IT/lbh-housing-register/commit/9607ca85e49f57438da87256a76516d7eedcca8d)

Wouldn't have guessed that this would work - quite unintuitive outcome given the nature of the error. But if it works, then yeah, let's go with it.

Duslerke (approved):

pages/applications/view/[id]/[person]/index.tsx
Comment (Duslerke) [reply pages/applications/view/[id]/[person]/index.tsx]: Found your problem

https://github.com/LBHackney-IT/lbh-housing-register/blob/4358cf5ad4ddaef374efa287e5c36d1a437f9b41/components/admin/HorizontalNav.tsx#L5

Comment (LBHTKarki) [reply pages/applications/view/[id]/[person]/index.tsx]: Yeah I tried both of those options:
1. Results in the same TS error
2. Results in the useless fragment linting error

Interesting, but the "wrapper" fragments results in seemengly valid TS and linter is OK too and the feature works, so I'll go with that for now.

Comment (Duslerke) [reply pages/applications/view/[id]/[person]/index.tsx]: Now it all makes sense. By wrapping the whole thing into fragment, you made it so your `children` that you pass into horrizontalnav is a single JSX element. Where as before it was an array of JSX elements, where some of the elements were `boolean | JSX.Element[]`.

Comment (Duslerke) [reply pages/applications/view/[id]/[person]/index.tsx]: At this point the solution is subjective. So I'll give another alternative to choose from:

You could tweak the type linked above to be:
```ts
children: JSX.Element | (JSX.Element | boolean)[];
```

Comment (Duslerke) [reply pages/applications/view/[id]/[person]/index.tsx]: This will make it explicit within the parent component's type declaration that some child components are rendered conditionally.

Comment (Duslerke) [reply pages/applications/view/[id]/[person]/index.tsx]: That way you should be able to drop the fragment wrapper.

Comment (LBHTKarki) [reply pages/applications/view/[id]/[person]/index.tsx]: Thank you! That's a lot tidier and makes it's more clear how you are supposed to handle the children. Updated heer https://github.com/LBHackney-IT/lbh-housing-register/pull/403/commits/f526803f3b3d9a1eebe4c3195423fd5668f2867c

PR #415, opened 2024-09-16
Ts 1666 hr update node and packages

sam-drumm:
https://hackney.atlassian.net/browse/TS-1666

**What**
- Updates node to 20LTS and packages to the latest possible without breaking changes,
- Updates Typescript to 4.9.5.

**Why**
- We are required to bump node from 14 as it's no longer supported in our build. It's also out of security support. 20LTS is the version that will keep us in support for the longest time and is possible to us,
- Likewise some packages require a later typescript, and it's recommended for this version of next so that's been increased to the latest in 4.x

**How**
- All packages have been upgradedto the latest versions that doesn't have breaking changes,
- As part of this I have also removed cypress-circleci-reporter and mocha as they are not required for our test results reporting,
- I have also removed mocks-server as it's no longer in use.


sentry.edge.config.js
```diff
@@ -0,0 +1,19 @@
...
+
+import * as Sentry from '@sentry/nextjs';
+
+const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN;
+const ENVIRONMENT = process.env.NEXT_PUBLIC_ENV;
+
+Sentry.init({
+  dsn:
+    SENTRY_DSN ||
+    'https://6fb0dd07e0fc4a75b0ab84b8e1f36460@o183917.ingest.sentry.io/6292602',
```
Duslerke, line 14: Just confirming.
This is the public URL, right?

Duslerke (approved):
Seems alright that I can see.

LBHTKarki (approved):

PR #421, opened 2024-09-26
TS-1659 fix person details null overwrite bug

LBHTKarki:
## WHAT
Current Lambda function on production is running on Node v14 runtime. That version of Node is now out of support and we cannot deploy changes using that runtime anymore. As part of the read only access feature development work we have updated the project to use the latest LTS Node version, 20.17. However that introduced some critical bugs that will stop us from deploying to production until they are fixed.

We noticed an issue where sometimes initial signup details where lost when user progressed from the login stage to the household members view. After troubleshooting the issue locally it appeared that the application does not handle delayed API calls properly which is causing the database and the application local state (using Redux) to get out of sync.

When the issue occurs the flow of events are as follows:
1. User fills in their personal details on the start page
2. User submits the details which causes API PATCH endpoint to be called while at the same time router push is used to move the user to the next page in the sign up flow i.e. agree-terms
3. When user agree the terms they are then "pushed" to the /apply/household page
4. When user gets to the household page they see undefined undefined in the main applicant section rather than their first and last name 

This behaviour was successfully created locally by delaying the API PATCH call, so it wasn't complete before the next page (agree-terms) was loaded.

The way the pages are designed to work is that start page updates the application details in the database on submit and then when user is pushed to the agree-terms page it loads the data from the database and uses that to update the local Redux store with the application details. This state is then used in turn in the switch from agree-terms to apply/household page to patch the application in the database again. In the above scenario the first patch that adds the person details to the application is not complete before the next get request occurs which means that the person is null in the local state and later calls.

This update addresses the issue by updating the start page to ensure the application has been updated in the database before pushing the user to the agree-terms page. This has been successfully tested locally and the person details now persists regardles of the delay in the application patch. 

## WHY
When the bug happens user ends up in unrecoverable situation where they cannot continue with the application.

## HOW
The following updates have been made to the application to resolve this issue:

1. Add new `hrApiCallsStatus` slice to the store, so we can track the status of various API calls. This slice only has extra reducers that update the new slice based on the `patchApplication` async thunk events
2. Update the `patchApplication` thunk to better handle the API call states and add support for handling failures
3. Update the start page to wait for the PATCH call to complete before pushing user to the next page
4. Update the error summary component for better testing support
5. Update Cypress config with new patch end point mock. Also minor tweaks to the get mock
6. Add test coverage for the store and page changes. The approach taken on testing is based on the recommendations from the Redux team i.e. majority of the tests should be integration tests where the whole application and components are tested together. For that reason the `patchApplication` and the slice & extra reducers are not tested individually, but as part of Cypress e2e test where the pages operate as they normally would. API calls are mocked at network level using nock which is also the recommended way to mock APIs when Redux is used

## FUTURE
1. Go through the rest of the pages to ensure they handle delayed API calls the same way to avoid any data loss 
2. Tidy up the API mock setups now that we are passing more and more props

### DIFF
--- a/cypress/e2e/pages/startApplicationLoggedIn.cy.ts
+++ b/cypress/e2e/pages/startApplicationLoggedIn.cy.ts
@@ -31,7 +31,7 @@ const applicationWithMainApplicant = {
   calculatedBedroomNeed: 1,
 };
 
-const fillInTheForm = () => {
+const fillInTheSignUpForm = () => {
   StartPage.getTitleDropdown().select(title);
   StartPage.getFirstNameInput().type(faker.person.firstName());
   StartPage.getLastNameInput().type(faker.person.lastName());
@@ -89,7 +89,7 @@ describe('Application', () => {
     StartPage.visit();
 
     //fill in the personal details form and submit
-    fillInTheForm();
+    fillInTheSignUpForm();
     StartPage.getSubmitButton().click();
 
     //check that message is shown until (delayed) PATCH call has finished
@@ -123,7 +123,7 @@ describe('Application', () => {
     StartPage.visit();
     StartPage.getErrorSummary().should('not.exist');
 
-    fillInTheForm();
+    fillInTheSignUpForm();
     StartPage.getSubmitButton().click();
 
     StartPage.getErrorSummary().should('be.visible');
--- a/cypress/e2e/pages/startApplicationLoggedIn.cy.ts
+++ b/cypress/e2e/pages/startApplicationLoggedIn.cy.ts
@@ -1,0 +1,132 @@
+import StartPage from '../../pages/start';
+import { generateApplication } from '../../../testUtils/applicationHelper';
+import {
+  generatePerson,
+  getRandomGender,
+  TitleEnum,
+} from '../../../testUtils/personHelper';
+import { faker } from '@faker-js/faker/locale/en_GB';
+import AgreeTermsPage from '../../pages/agreeTerms';
+import { StatusCodes } from 'http-status-codes';
+
+const applicationId = faker.string.uuid();
+const personId = faker.string.uuid();
+const title = faker.helpers.enumValue(TitleEnum);
+const birthDate = faker.date.birthdate({ mode: 'age', min: 18, max: 70 });
+const mainApplicant = generatePerson(personId);
+const apiResponseDelay = 1000;
+const application = generateApplication(applicationId, personId, false, false);
+const phoneNumber = faker.phone.number();
+
+const applicationWithMainApplicant = {
+  ...application,
+  mainApplicant: {
+    ...application.mainApplicant,
+    contactInformation: {
+      ...application.mainApplicant.contactInformation,
+      phoneNumber: phoneNumber,
+    },
+    person: mainApplicant,
+  },
+  calculatedBedroomNeed: 1,
+};
+
+const fillInTheForm = () => {
+  StartPage.getTitleDropdown().select(title);
+  StartPage.getFirstNameInput().type(faker.person.firstName());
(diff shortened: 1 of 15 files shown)


pages/apply/agree-terms.tsx
```diff
@@ -24,7 +24,7 @@ const ApplicationTermsPage = (): JSX.Element => {
   };
 
   return (
-    <Layout pageName="Agreement">
+    <Layout pageName="Agreement" dataTestId="agree-terms-page">
```
sam-drumm, line 27: semantic but we have generally prefixed dataTestId with test - may be worthwhile for consistency at least.

pages/apply/start.tsx
```diff
@@ -7,18 +7,41 @@ import { HeadingOne } from '../../components/content/headings';
...
+
+  useEffect(() => {
+    if (
+      patchApplicationStatus?.callStatus === ApiCallStatusCode.FULFILLED &&
+      hasSignedUp
+    ) {
+      router.push('/apply/agree-terms');
+    }
+
+    if (patchApplicationStatus?.callStatus == ApiCallStatusCode.REJECTED) {
```
sam-drumm, line 38: Should this be strict ===

cypress/e2e/pages/startApplicationLoggedIn.cy.ts
sam-drumm, line 1: I think this is one for later, but it might be worth us reviewing the naming in cypress. I.e. we have cy/e2e/pages and cy/pages. Both make sense independently but could be worth a review.

cypress/e2e/pages/startApplicationLoggedIn.cy.ts
```diff
@@ -0,0 +1,132 @@
...
+};
+
+describe('Application', () => {
+  beforeEach(() => {
+    cy.clearAllCookies();
+    cy.task('clearNock');
+  });
+
+  it('shows the loading spinner while application data is being fetched on the background', () => {
+    console.dir(application);
```
sam-drumm, line 55: some logs

lib/store/apiCallsStatus.ts
```diff
@@ -0,0 +1,64 @@
...
+        };
+      })
+      .addCase(updateApplication.fulfilled, (state) => {
+        state.patchApplication = {
+          callStatus: ApiCallStatusCode.FULFILLED,
+          error: null,
+        };
+      })
+      .addCase(updateApplication.rejected, (state, action) => {
+        console.dir(action.payload);
```
sam-drumm, line 51: some logs

cypress/e2e/pages/startApplicationLoggedIn.cy.ts
LBHTKarki: I reckon we can leave this for now and tidy it up in a follow up PR

lib/store/apiCallsStatus.ts
LBHTKarki: removed in https://github.com/LBHackney-IT/lbh-housing-register/pull/421/commits/0f9db61943508002de2b659ea834336a9978955c

pages/apply/agree-terms.tsx
LBHTKarki: yes, definitely. Updated in https://github.com/LBHackney-IT/lbh-housing-register/pull/421/commits/0f9db61943508002de2b659ea834336a9978955c

pages/apply/start.tsx
LBHTKarki: yes 👍 , updated in https://github.com/LBHackney-IT/lbh-housing-register/pull/421/commits/0f9db61943508002de2b659ea834336a9978955c

sam-drumm (approved):

PR #422, opened 2024-09-27
General bug fixes

sam-drumm:
**What**
- Uses a uniqueID for lists rather than index or non-unique value,
- Uses a forEach method where a returned array is not required (only a side effect),
- Uses optional chaining rather than && for simplicity,
- Adds a bundle analyser so that we can investigate opportunities to reduce unused JS,
- Visual change to the Footer component to manage an empty id (visually) see screenshot,
- Updates deprecated circleci python image,
- Update aws-cli to a more modern image,
- Updates configuration with GTAG (also in param store). Analytics team have setup a new tag under the hackney umbrella,
- Only stores Cypress videos on failed tests and compresses them. 

**Why**
- Responding to flags from Sonar relating to maintainability,
- Page performance is poor, so at some point in the future we want to investigate this.

Screenshot 2024-09-26 at 12 45 36
Screenshot 2024-09-26 at 12 45 45
Screenshot 2024-09-26 at 12 56 19
Screenshot 2024-09-26 at 12 56 37


lib/utils/resident.ts
```diff
@@ -45,7 +45,7 @@ export const getApplicationSectionsForResident = (
   if (isMainApplicant) {
     return getMainApplicantQuestions();
   } else {
-    return getOtherMemberQuestions(isOver18, isPartner);
```
sam-drumm, line 48: Matching the order of the function here.

components/admin/AddCaseAddress.tsx
```diff
@@ -141,7 +143,7 @@ export default function AddCaseAddress({
           <SummaryListActions wideActions={true}>
             {addresses
               ? addresses.map((addressItem, index) => (
-                  <FormGroup key={index}>
+                  <FormGroup key={uniqueID()}>
```
Duslerke, line 146: Introducing an extra NPM library for this is overkill, you could just do:
```ts
addresses.map((addressItem, index) => (
                  <FormGroup key={`address-item-${index}`}>
                     ...
```
and that should be unique enough. :thinking:

pages/applications/view/[id]/[person]/index.tsx
```diff
@@ -91,10 +91,8 @@ export default function ApplicationPersonPage({
   )}%20${cleanUpParams(applicant?.person?.surname ?? '')}`;
 
   return (
-    // eslint-disable-next-line react/jsx-no-useless-fragment
     <>
-      {data.id ? (
-        // eslint-disable-next-line react/jsx-no-constructed-context-values
+      {data && data.id ? (
```
Comment (Duslerke) [inline pages/applications/view/[id]/[person]/index.tsx:95]: Could probably do:

```ts
data?.id ? ( ... )
```

Duslerke (commented):
The changes are repetitive, so I left the comments only next to first instance of a case.

Changes look good.
Though, I think nice-to-have extra libraries like `uuid` in this case should be avoided with all the supply chain attacks these days.

components/admin/AddCaseAddress.tsx
sam-drumm: I've gone unique (without a huge amount of consideration to each case) is it's the recommended approach in react rather than using array index that might change/reorder be filtered etc.

Take your point on avoiding adding another package - I might take a look and see if any of the packages we currently use have a generator. But uuid pretty solid package.

Duslerke: Unless your elements change order between re-renders, you probably don't need a random key,
but if you do want a random one anyway, then you don't need any package in general - all you need should be in the base node.js already as part of Math library:

```js
const randKey = () => Math.floor(Math.random() * 10e6)
```

sam-drumm: Yeah, agree it's not strictly necessary in many cases, but it's probably good practice and a bit safer. 

Re using built in math - this generally seems pretty discouraged / seen as an unstable approach.

Duslerke: Seems like the react doesn't recommend any randomness in the keys.
They only say you should use `uuid` if your items or data you're referring to is persisted .. to say local storage.

| image |
| --- |
| Snippet from react docs |

sam-drumm: good shout - I needed to read the documentation properly. To be fair, a tonne of conflicting stuff around the approach. But basically the ID should stable, irrelevant of re-renders, which makes sense. 

Will go back through this for a tidy, but it seems like in most cases the approach will have to be the index plus a relevant value. Thanks :)

Duslerke (approved):

LBHTKarki (approved):

PR #426, opened 2024-09-30
add tests for read only notes section

1JW1:
## What

Added a test to check that the notes section is not visible in read only mode

## Why 

To improve the test coverage of the housing register application


cypress/e2e/pages/addNotesReadOnly.cy.ts
```diff
@@ -0,0 +1,26 @@
...
+    ReviewApplicantPage.mockHousingRegisterApiGetApplications(
+      applicationId,
+      application
+    );
+  });
+
+  it("doesn't show the notes section for read only users", () => {
+    ReviewApplicantPage.visit(applicationId, personId);
+    ReviewApplicantPage.getNotesSection().should('not.exist');
+  });
```
Duslerke, line 25: Nice to see such tests added.
Though, you should cover both bad and happy paths.

In this case you'd need a test that tests whether notes are shown for other users that are meant to see them.

PR #427, opened 2024-10-01
Adds tests for agree terms

sam-drumm:
https://hackney.atlassian.net/browse/TS-1690?atlOrigin=eyJpIjoiNjM3MDBlZjg5OTRlNDc5MjhhMjVhZDVhYzBhZDJmMjgiLCJwIjoiaiJ9

**What**
- Implements the new slice on the agree conditions page to make sure patch is fulfilled before progressing to the next page
- Creates a custom hook to make this logic easier to repeat in other places.

**Why**
- This is a further implementation of the solution to the bug [detailed here](https://github.com/LBHackney-IT/lbh-housing-register/pull/421).

### DIFF
--- a/lib/store/apiCallsStatus.ts
+++ b/lib/store/apiCallsStatus.ts
@@ -25,10 +25,6 @@ const initialState: HRApiCallsStatusState = {
   },
 };
 
-export const selectPatchApplicationStatus = (
-  state: RootState
-): ApiCallStatus | undefined => state.hrApiCallsStatus.patchApplication;
-
 export const selectSaveApplicationStatus = (
   state: RootState
 ): ApiCallStatus | undefined => state.hrApiCallsStatus.patchApplication;
--- a/lib/store/apiCallsStatus.ts
+++ b/lib/store/apiCallsStatus.ts
@@ -29,6 +29,10 @@ export const selectPatchApplicationStatus = (
   state: RootState
 ): ApiCallStatus | undefined => state.hrApiCallsStatus.patchApplication;
 
+export const selectSaveApplicationStatus = (
+  state: RootState
+): ApiCallStatus | undefined => state.hrApiCallsStatus.patchApplication;
+
 export const hrApiCallsStatus = createSlice({
   name: 'hrApiCallsStatus',
   initialState,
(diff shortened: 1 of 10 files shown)


lib/store/apiCallsStatus.ts
```diff
@@ -29,6 +29,10 @@ export const selectPatchApplicationStatus = (
   state: RootState
 ): ApiCallStatus | undefined => state.hrApiCallsStatus.patchApplication;
 
+export const selectSaveApplicationStatus = (
```
LBHTKarki, line 32: Is this same as the existing `selectPatchApplicationStatus` selector? Perhaps we can drop one of them ?

LBHTKarki: save is probably more appropriate than patch in the name

pages/apply/agree-terms.tsx
```diff
@@ -1,31 +1,65 @@
+import {
+  // useEffect,
```
LBHTKarki, line 2: commnet

pages/apply/agree-terms.tsx
```diff
@@ -1,31 +1,65 @@
...
+} from 'react';
 import { HeadingOne } from '../../components/content/headings';
 import Paragraph from '../../components/content/paragraph';
 import Layout from '../../components/layout/resident-layout';
 import { useAppDispatch, useAppSelector } from '../../lib/store/hooks';
 import { agree } from '../../lib/store/mainApplicant';
 import { getFormData, FormID } from '../../lib/utils/form-data';
 import Form from '../../components/form/form';
-import router from 'next/router';
+// import router from 'next/router';
```
LBHTKarki, line 12: comment

pages/apply/agree-terms.tsx
```diff
@@ -1,31 +1,65 @@
...
+import { Errors } from 'lib/types/errors';
+import Loading from 'components/loading';
 
 const ApplicationTermsPage = (): JSX.Element => {
   // TODO: might not be right place for this,
   // but we need to ensure new user is linked to application
   const dispatch = useAppDispatch();
   const applicationId = useAppSelector((store) => store.application.id);
+  const [hasSaved, setHasSaved] = useState<boolean>(false);
+  const patchApplicationStatus = useAppSelector(selectPatchApplicationStatus);
```
LBHTKarki, line 30: We can review the naming of this. Save rather than patch perhaps?

LBHTKarki (approved):

PR #428, opened 2024-10-03
TS-1689 improve error handling on signin and verify pages

LBHTKarki:
## WHAT
Currently `sign-in` and `verify` pages don't handle Housing Register API errors as well as they should. Also tidies up some other unrelated components, for example replaces Cypress intercepts with nock based mocks.

## WHY
Improves the error handling, so user is aware at all times if any of the API calls fail.

## HOW
1. Update make an application test to use nock based API mocks rather than Cypress intercepts. This is so we are using consistent mocks setups in test.
2. Make some existing error messages a little bit less revealing
3. Add test coverage to sign in and verify pages to cover the updated error handling
4. Rename some of the cypress test pages to better match what they are used for
5. Add new mocks for generate and verify token end points
6. Remove all existing cypress intercept setups. They are no longer needed
7. Rename the useApplicationUpdateStatus hook to make it more generic
8. Rename the patchApplication property in hrApiCallsStatus slice to updateApplication so it matches the name of the related async thunk
9. Update `createVerifyCode` and `confirmVerifyCode` thunks with better error handling
10. Add new helper for genrating email addresses


that-Gui (approved):
LGTM 👍🏻

cypress/e2e/pages/agreeTerms.cy.ts
```diff
@@ -104,11 +104,10 @@ describe('Application', () => {
       applicationWithMainApplicant
     );
 
-    const errorStatus = 'Bad Request';
     const errorStatusCode = StatusCodes.BAD_REQUEST;
 
     //based on current API layer setup
-    const expectedErrorMessage = `Unable to update application: ${errorStatus} (${errorStatusCode})`;
+    const expectedErrorMessage = `Unable to update application (${errorStatusCode})`;
```
sam-drumm, line 110: Just from a FE perspective, I wonder if it would be worth having a generic message here i.e.

If you need help with this page please contact.  

Just thinking it very much like sorry - bad luck.

sam-drumm (approved):
Looks good - one small comment which could be something to do in a cleanup.

PR #431, opened 2024-10-03
TS-1703 Improve data and error handling on edit person page

LBHTKarki:
## WHAT
`/applications/edit/[id]/[person]/index.tsx ` page is not currently displaying and handling HR API errors correctly. This update adds better saving and error handling messages, so user is aware what's happening.

## WHY
User don't currently get enough feedback on what's happening on the page when data is saved.

## HOW
1. Show saving... message to user when applications is saved 
2. Show error message when HR API call fails. This had to be added to the `MainApplicantForm` component too due to way the page is currently structured
3. Update the gateway method by removing the current status === 400 check and use generic error instead. Object returned by the API layer when status is 400 does not include message property, so that will always return undefined message. It's better to just show generic error and status code for now. 

## FUTURE
Build proper error handling for 400 errors happening at the backend, so that they can be shown to users in a friendly format. Right now we rely on the front end validation and if user gets passed that we have to look at the errors in the conesole. This is internal page only, so we can update this later.

### DIFF
--- a/components/admin/MainApplicantForm.tsx
+++ b/components/admin/MainApplicantForm.tsx
@@ -1,4 +1,4 @@
-import { Form, Formik, FormikValues, FormikErrors } from 'formik';
+import { Form, Formik, FormikValues } from 'formik';
 import { UserContext } from '../../lib/contexts/user-context';
 import Button from '../../components/button';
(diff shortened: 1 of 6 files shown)


LBHTKarki:
I'll tweak the error handling a bit more before creating a PR

PR #432, opened 2024-10-07
TS-1703 Improve data and error handling on edit application pages

LBHTKarki:
## WHAT
Edit application related pages are not currently displaying and handling HR API errors correctly. This update adds better saving and error handling messages, so user is aware what's happening. Pages included:
`/applications/edit/[id]/[person]/index.tsx ` 
`/applications/edit/[id]/[person]/edit-household-member.tsx ` 
`/applications/edit/[id]/add-household-member.tsx ` 

## WHY
User don't currently get enough feedback on what's happening on the page when data is saved or errors occur.

## HOW
1. Show `Saving...` message to user when applications is saved 
2. Show error message when HR API call fails. This had to be added to the `MainApplicantForm` and `HouseholdMemberForm` components too due to way the pages are currently structured
3. Update the gateway method by removing the current status === 400 check and use generic error instead. Object returned by the API layer when status is 400 does not include message property, so that will always return undefined message. It's better to just show generic error and status code for now until we have a chance to tidy up the 400 errors properly. 
4. es-lint errors have been disabled for parts of the code that are not affected by these changes

## FUTURE
Build proper error handling for 400 errors happening at the backend, so that they can be shown to users in a friendly format. Right now we rely on the front end validation and if user gets passed that we have to look at the errors in the conesole. These are internal pages only, so we can update them later to be more user friendly.

### DIFF
--- a/components/admin/HouseholdMemberForm.tsx
+++ b/components/admin/HouseholdMemberForm.tsx
@@ -1,4 +1,4 @@
-import { Form, Formik, FormikValues, FormikErrors } from 'formik';
+import { Form, Formik, FormikValues } from 'formik';
 import { UserContext } from '../../lib/contexts/user-context';
 import Button from '../../components/button';
 import ErrorSummary from '../../components/errors/error-summary';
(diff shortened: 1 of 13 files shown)


sam-drumm (approved):
All looks good - tests all passing.

PR #434, opened 2024-10-08
TS-1702: adds error handling on apply pages

sam-drumm:
https://hackney.atlassian.net/browse/TS-1702?atlOrigin=eyJpIjoiNWU5OGVmZDE4YzM4NGYwN2IyZjQ0OWQyODJmYjU5ZGUiLCJwIjoiaiJ9

**What**
Ensure that promises are fulfilled before progressing on the apply/ pages

**Why**
This is an implementation of the original issue detailed https://github.com/LBHackney-IT/lbh-housing-register/pull/421

**How**
Where the action triggers middleware to save, the custom hook is implemented to keep track of the promise state.

Where the action triggers a thunk directly, error handling is added to the page and the thunk is adapted to return an error on unfulfilled promise.

### DIFF
--- a/cypress/pages/ethnicity.ts
+++ b/cypress/pages/ethnicity.ts
@@ -1,0 +1,27 @@
+class EthnicityPage {
+  static getEthnicityPage() {
+    const testId = 'test-ethnicity-questions-page';
+    return cy.get(`[data-testid="${testId}"]`);
+  }
+  static visit() {
+    cy.visit(`/apply/submit/ethnicity-questions`);
+  }
+  // static getEthnicityRadios() {
+  //   const testId = 'test-radio-';
+  //   return cy.get(`[data-testid*="${testId}"]`);
+  // }
+  // static getSaveButton() {
+  //   const testId = 'test-submit-form-button';
+  //   return cy.get(`[data-testid="${testId}"]`);
+  // }
+  static getErrorSummary() {
+    const testId = 'test-ethnicity-questions-error-summary';
+    return cy.get(`[data-testid="${testId}"]`);
+  }
+  // static getLoadingSpinner() {
+  //   const testId = 'test-loading-spinner';
+  //   return cy.get(`[data-testid="${testId}"]`);
+  // }
+}
+
+export default EthnicityPage;
--- a/cypress/pages/ethnicity.ts
+++ b/cypress/pages/ethnicity.ts
@@ -6,22 +6,11 @@ class EthnicityPage {
   static visit() {
     cy.visit(`/apply/submit/ethnicity-questions`);
   }
-  // static getEthnicityRadios() {
-  //   const testId = 'test-radio-';
-  //   return cy.get(`[data-testid*="${testId}"]`);
-  // }
-  // static getSaveButton() {
-  //   const testId = 'test-submit-form-button';
-  //   return cy.get(`[data-testid="${testId}"]`);
-  // }
+
   static getErrorSummary() {
     const testId = 'test-ethnicity-questions-error-summary';
     return cy.get(`[data-testid="${testId}"]`);
   }
-  // static getLoadingSpinner() {
-  //   const testId = 'test-loading-spinner';
-  //   return cy.get(`[data-testid="${testId}"]`);
-  // }
 }
 
 export default EthnicityPage;
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "test": "jest && tsc --noEmit",
     "test:watch": "jest --watch",
     "prepare": "husky",
-    "cypress:open": "cypress open --browser electron",
+    "cypress:open": "cypress open --browser chrome",
     "e2e:run": "cypress run --e2e --browser electron",
     "component:run": "cypress run --component --browser electron"
   },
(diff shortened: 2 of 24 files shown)


cypress/pages/ethnicity.ts
```diff
@@ -0,0 +1,27 @@
+class EthnicityPage {
+  static getEthnicityPage() {
+    const testId = 'test-ethnicity-questions-page';
+    return cy.get(`[data-testid="${testId}"]`);
+  }
+  static visit() {
+    cy.visit(`/apply/submit/ethnicity-questions`);
+  }
+  // static getEthnicityRadios() {
```
LBHTKarki, line 9: comments

cypress/pages/ethnicity.ts
```diff
@@ -0,0 +1,27 @@
...
+  // }
+  // static getSaveButton() {
+  //   const testId = 'test-submit-form-button';
+  //   return cy.get(`[data-testid="${testId}"]`);
+  // }
+  static getErrorSummary() {
+    const testId = 'test-ethnicity-questions-error-summary';
+    return cy.get(`[data-testid="${testId}"]`);
+  }
+  // static getLoadingSpinner() {
```
LBHTKarki, line 21: comments

package.json
```diff
@@ -14,7 +14,7 @@
     "test": "jest && tsc --noEmit",
     "test:watch": "jest --watch",
     "prepare": "husky",
-    "cypress:open": "cypress open --browser electron",
+    "cypress:open": "cypress open --browser chrome",
```
LBHTKarki, line 17: is this on purpose?

pages/apply/submit/declaration.tsx
```diff
@@ -27,72 +31,108 @@ const Declaration = (): JSX.Element => {
...
         getDisqualificationReasonOption(reason)
       );
       const reason = reasonStrings.join(',');
+      // this is just a notification, so for now we don't need to handle errors
       dispatch(sendDisqualifyEmail({ application, reason }));
-      dispatch(disqualifyApplication(application.id!));
-      router.push('/apply/not-eligible');
+
+      try {
+        console.log('hello');
```
LBHTKarki, line 50: console.log

pages/apply/submit/declaration.tsx
```diff
@@ -27,72 +31,108 @@ const Declaration = (): JSX.Element => {
...
+        await dispatch(disqualifyApplication(application.id!)).unwrap();
```
LBHTKarki, line 51: `.unwrap()` here probably doesn't do anything unless chained with `then` and `catch`. Should the `router.push` happen inside `then` and errors handled in `catch` should the dispatch result in rejected action?

pages/apply/submit/declaration.tsx
```diff
@@ -27,72 +31,108 @@ const Declaration = (): JSX.Element => {
...
+        router.push('/apply/not-eligible');
+      } catch (error) {
+        setUserError(error as string);
```
LBHTKarki, line 54: should `setLoading(false);` be here too?

pages/apply/submit/declaration.tsx
```diff
@@ -27,72 +31,108 @@ const Declaration = (): JSX.Element => {
...
       if (medicalNeeds > 0) {
+        // these are just notifications, so we don't need to handle errors yet
         dispatch(sendMedicalNeed({ application, medicalNeeds }));
       }
-
-      dispatch(completeApplication(application));
-      dispatch(createEvidenceRequest(application));
-      router.push('/apply/confirmation');
+      try {
+        await dispatch(completeApplication(application)).unwrap();
```
LBHTKarki, line 70: same as above, would be good to review the setup when `unwrap()` is used.

pages/apply/submit/declaration.tsx
```diff
@@ -27,72 +31,108 @@ const Declaration = (): JSX.Element => {
...
+        await dispatch(createEvidenceRequest(application)).unwrap();
+        router.push('/apply/confirmation');
+      } catch (error) {
+        console.log('Error completing the application:', error);
```
LBHTKarki, line 74: `setLoading(false);` here too?

pages/apply/submit/ethnicity-questions.tsx
```diff
@@ -1,5 +1,5 @@
 import { FormikValues, getIn } from 'formik';
-import { useRouter } from 'next/router';
+// import { useRouter } from 'next/router';
```
LBHTKarki, line 2: comment

pages/apply/submit/declaration.tsx
LBHTKarki: looks like unwrap() will cause an error to be thrown in this case, so to say it wouldn't do anything was not correct

pages/apply/submit/declaration.tsx
sam-drumm: the finally should set the loading state back to false in the try or catch

package.json
sam-drumm: Yeah - seems more sensible to be loading with chrome with the redux dev-tools.

LBHTKarki (approved):
👍

PR #435, opened 2024-10-09
TS-1712 Improve error handling on add-case page

LBHTKarki:
## WHAT
Currently users are not notified of any backend errors on `/applications/add-case` page. This is the page where officers can create new cases directly from the app without resident having to create one of their own first. This update improves the error handling by showing an error message if any of the API calls made during the process fail. Saving message has also been added to indicate that the application creation process is in progress after form submission. The process includes the following steps in the backend:
1. Create the application
2. Complete the application
3. Update the application status to `MANUAL_DRAFT` 

## WHY
Users should get better feedback what's happening during the process. By adding error handling and saving status it's clear at all times what's happening in the backend.

## HOW
1. Update the affected end points to return error status code on API call failure
2. Update the `add-case` page to show those errors to the user
3. Add saving message to indicate that the  application creation process is currently running

## FUTURE

### DIFF
--- a/lib/gateways/internal-api.ts
+++ b/lib/gateways/internal-api.ts
@@ -27,7 +27,12 @@ export const createApplication = async (application: Application) => {
     method: 'POST',
     body: JSON.stringify(application),
   });
-  return (await res.json()) as Application;
+
+  if (res.ok) {
+    return (await res.json()) as Application;
+  } else {
+    throw Error(`Unable to create application (${res.status})`);
+  }
 };
 
 export const completeApplication = async (application: Application) => {
@@ -35,7 +40,11 @@ export const completeApplication = async (application: Application) => {
     method: 'PATCH',
     body: JSON.stringify(application),
   });
-  return (await res.json()) as Application;
+  if (res.ok) {
+    return (await res.json()) as Application;
+  } else {
+    throw Error(`Unable to complete application (${res.status})`);
+  }
 };
 
 export const generateNovaletExport = async () => {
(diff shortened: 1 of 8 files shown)


pages/applications/add-case.tsx
```diff
@@ -99,6 +115,9 @@ export default function AddCasePage({ user }: PageProps): JSX.Element {
       handleSaveApplication={handleSaveApplication}
       ethnicity={ethnicity}
       setEthnicity={setEthnicity}
+      dataTestId="test-add-case-page"
+      isSaving={isSaving}
+      userError={userError}
```
Duslerke, line 120: This component does not appear to have the `isSaving` and `userError` keys:
https://github.com/LBHackney-IT/lbh-housing-register/blob/e3b81f4691ec36c1e700b71a0d7ab33d79e7fc2b/components/admin/MainApplicantForm.tsx#L95-L106

sam-drumm (approved):
Same implementations - looks good. Have run tests locally.

lib/gateways/internal-api.ts
```diff
@@ -27,15 +27,24 @@ export const createApplication = async (application: Application) => {
     method: 'POST',
     body: JSON.stringify(application),
   });
-  return (await res.json()) as Application;
+
```
sam-drumm, line 30: Was this implemented by a previous PR? Guessing just out of date development at the time.

pages/applications/add-case.tsx
LBHTKarki: They are not in main yet, but are in this feature branch and also in development which this branch is based on. Are you perhaps looking at the component in main?

Duslerke: You're right. I was looking into main, my bad.
Didn't notice we have a `development` branch on this repository.

Duslerke (approved):

PR #442, opened 2024-10-21
Delay loading state

sam-drumm:
https://hackney.atlassian.net/browse/TS-1735?atlOrigin=eyJpIjoiYzczNjI1NWI3MGY2NDAxYThlNTBiMjJhYzNmMzZiMDUiLCJwIjoiaiJ9

**What**
- Moves the loading state management to the custom hook, 
- Allows an optional timeout to be set, where a visual delay could be useful before showing a loading text/spinner. 

**Why**
- Cleaner implementation that allows a pending state to be used in the page/component without getting the value,
- To avoid flashing, in some components/pages it can be beneficial to delay the loading state. For that reason, the hook can manage that by having a timeout value passed to it.

### DIFF
--- a/lib/hooks/useApiCallStatus.tsx
+++ b/lib/hooks/useApiCallStatus.tsx
@@ -12,6 +12,7 @@ interface UseApiCallStatusProps {
   query?: ParsedUrlQueryInput;
   setUserError: (error: string) => void;
   scrollToError: () => void;
+  pendingStatusStateDelay?: number;
 }
 
 const useApiCallStatus = ({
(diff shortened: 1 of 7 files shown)


lib/hooks/useApiCallStatus.tsx
```diff
@@ -21,10 +22,22 @@ const useApiCallStatus = ({
...
 }: UseApiCallStatusProps) => {
   const router = useRouter();
+  const [pendingStatus, setPendingStatus] = useState<boolean>(false);
+  let timer: NodeJS.Timeout;
 
   useEffect(() => {
+    if (selector?.callStatus === ApiCallStatusCode.PENDING) {
+      timer = setTimeout(() => {
+        setPendingStatus(true);
+      }, pendingStatusStateDelay);
```
Duslerke, line 35: Any chance of this launching multiple setTimeouts due to re-renders?

Igf so, perhaps, it would help to include the `pendingStatus` in the conditional to prevent it from happening multiple times?

```js
if (!pendingStatus && selector?.callStatus === ApiCallStatusCode.PENDING) { ... }
```

Duslerke (approved):

LBHTKarki (approved):

lib/hooks/useApiCallStatus.tsx
sam-drumm: Yeah that's a good point thanks.

PR #443, opened 2024-10-23
Fixes the missing value prop

sam-drumm:
https://hackney.atlassian.net/browse/TS-1736?atlOrigin=eyJpIjoiNGM2ZmMyNmU0Y2FhNDY3N2JkNTNlMjdhNDkwNzY4YjQiLCJwIjoiaiJ9

**What**
The input field was being passed a conditionalDisplay prop and throwing a warning. This PR cleans the prop to eliminate the warning. 

Screenshot 2024-10-23 at 09 16 56

### DIFF
--- a/components/form/input.tsx
+++ b/components/form/input.tsx
@@ -24,6 +24,12 @@ export default function Input({
   hideLabel,
   ...additionalInputProps
 }: InputProps): JSX.Element {
+  const { conditionalDisplay, ...otherValues } = additionalInputProps;
+  const additionalInputPropsWithConditionalDisplay = {
+    ...otherValues,
+    conditionaldisplay: conditionalDisplay,
+  };
+
   return (
     <div>
       <Field name={name}>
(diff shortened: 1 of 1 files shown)


components/form/input.tsx
```diff
@@ -24,6 +24,12 @@ export default function Input({
   hideLabel,
   ...additionalInputProps
 }: InputProps): JSX.Element {
+  const { conditionalDisplay, ...otherValues } = additionalInputProps;
+  const additionalInputPropsWithConditionalDisplay = {
+    ...otherValues,
+    conditionaldisplay: conditionalDisplay,
+  };
+
```
Duslerke, line 32: What's the difference here?

Let's say `additionalInputProps` is a set with items A, B, C.
We destructure it to: `{ A, theRest }`
Then recreate the object to be a set that's: `{ theRest, A }`, which is just: `{B, C, A}`

`additionalInputProps` and `additionalInputPropsWithConditionalDisplay` have the same keys.

Duslerke: Oh, it's just the casing change of `D` to `d`?

Duslerke: But then why not pass it to the input as a (lowercased) `conditionaldisplay` from the get go?

sam-drumm: yeah exactly - it's just to get rid of the warning without cleaning up the whole thing.

Duslerke (approved):

LBHTKarki (approved):

PR #448, opened 2024-10-24
TS-1575 sample local e2e test

LBHTKarki:
Sample e2e test that fills in the entire form. This is designed to be run against yet undocumented local test setup where we run production FE build against real local backend. This is meant to give us an environment that behaves very closely to real environment and will aid in preparation to UAT.


sam-drumm (approved):

cypress/e2e/local/sampleForSingleEligibleApplicant.cy.ts
```diff
@@ -0,0 +1,258 @@
...
+        postcode: `${postcode}`,
+        UPRN: 11111111111,
+      },
+    ],
+    page_count: 1,
+    total_count: 1,
+  },
+};
+
+Cypress._.times(1, () => {
```
Duslerke, line 68: Never seen this before, what is this for?

Couldn't we just start with the `describe` block?

Duslerke (approved):

cypress/e2e/local/sampleForSingleEligibleApplicant.cy.ts
LBHTKarki: This way you can run the tests in that block as many times as you like back to back. Sometimes handy for troubleshooting if things seem a bit flaky.

PR #450, opened 2024-10-25
Ts 1756 hr pipeline is broken due to nock config errors

sam-drumm:
https://hackney.atlassian.net/browse/TS-1756?atlOrigin=eyJpIjoiODJhYjYxZDdkMjg1NDI4YzkwNWFkMDI2MTFiZWRiZWUiLCJwIjoiaiJ9

**What**
Allows the pipeline to run without errors unrelated to the tests or code. 

- Disables Firefox in tests due to the chunk error. Some investigation of options to turn this back on has been done and can be completed at a later date. [See ticket](https://hackney.atlassian.net/browse/TS-1770?atlOrigin=eyJpIjoiOGNiYmZmNmFjNjE1NGY4ZGI3YThhMjAyZmZlNjc2ZDMiLCJwIjoiaiJ9)
- Disables the webpack bundle analyser. As this was triggering on each test run, a solution should be found to only run in development mode and not during testing. 
- Excludes tests in the local folder from running in the pipeline. The cypress.config will exclude the folder if the PIPELINE variable is set.
- Re-enables video and video compression as errors are FF specific. The config will only save a video on failure, and compression means videos are about 20% of previous size at about 120kb. 

**Why**
- Our priority to test the new features, so any config enhancements can be cleaned up at a later stage.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,7 +95,7 @@ jobs:
           install-command: npm install
           post-install: npm run build
       - cypress/run-tests:
-          cypress-command: npx cypress run --browser=<< parameters.browser >> --config viewportHeight=<< parameters.height >>,viewportWidth=<< parameters.width >> --reporter junit --reporter-options "mochaFile=test_results/cypress-e2e-[hash].xml --spec "!.cypress/e2e/local/**/*"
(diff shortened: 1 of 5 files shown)


martapederiva (approved):
:shipit:

LBHTKarki (approved):

Duslerke (approved):
Sad to see the firefox go.
What was the issue with it?

sam-drumm:
@Duslerke It won't defeat me - a temporary measure 😅. It's basically firefox doesn't allow video recording without some specific interactions first, so it's causing some flow on problems. Its solvable, but probably better to focus on writing the tests themselves first.

PR #451, opened 2024-10-25
TS-1769 local e2e eligible single over 55

LBHTKarki:
Local e2e tests for the following scenarios:
1.  Add household members to the household at the beginning of the application
2. Eligible main applicant only (over 55 years old)
3. Eligible application with household members

Also fixes the sample test flow.

### DIFF
--- a/cypress/e2e/local/eligibleMainApplicantOnlyOver55.cy.ts
+++ b/cypress/e2e/local/eligibleMainApplicantOnlyOver55.cy.ts
@@ -81,7 +81,7 @@ Cypress._.times(1, () => {
         addressSearchAPIResponse
       ).as('addressSearchMock');
 
-      cy.viewport(1000, 1000);
+      cy.viewport(1280, 720);
 
       HomePage.visit(applicationId);
       HomePage.getCookiesButton().click();
--- a/cypress/e2e/local/eligibleMainApplicantOnlyOver55.cy.ts
+++ b/cypress/e2e/local/eligibleMainApplicantOnlyOver55.cy.ts
@@ -110,17 +110,24 @@ Cypress._.times(1, () => {
       AgreeTermsPage.getAgreeCheckbox().check();
       AgreeTermsPage.getAgreeButton().click();
 
+      cy.contains(`${mainApplicantFirstName} ${mainApplicantLastName}`);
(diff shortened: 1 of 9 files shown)


martapederiva (approved):
🚢

cypress/pages/apply/[resident]/your-situation.ts
```diff
@@ -3,8 +3,8 @@ class ApplyResidentYourSituationPage {
     cy.visit(`/apply/${personId}/your-situation`);
   }
 
-  static getServedInArmedForcesRadioButton() {
-    const testId = 'test-radio-situation-armed-forces.1';
+  static getServedInArmedForcesRadioButton(index: number) {
+    const testId = `test-radio-situation-armed-forces.${index}`;
```
Comment (Duslerke) [inline cypress/pages/apply/[resident]/your-situation.ts:7]: Never seen a dot being used in a selector name.
Would guess it's not exactly conventional, but probably not an issue.

cypress/e2e/local/eligibleMainApplicantOnlyOver55.cy.ts
```diff
@@ -0,0 +1,264 @@
...
+      //intercept address API since address is not relevant in this test
+      cy.intercept(
+        {
+          method: 'GET',
+          path: '/api/address/*',
+        },
+        addressSearchAPIResponse
+      ).as('addressSearchMock');
+
+      cy.viewport(1000, 1000);
```
Duslerke, line 84: How come a square view port?

Duslerke (approved):

cypress/pages/apply/[resident]/your-situation.ts
Comment (LBHTKarki) [reply cypress/pages/apply/[resident]/your-situation.ts:7]: These controls use index based numbering in the id with `.`, so these test IDs follow the same convention. I agree that it's a bit unusual, but we can tidy those up later.

cypress/e2e/local/eligibleMainApplicantOnlyOver55.cy.ts
LBHTKarki: It's just a random number to bump up the default viewport size a bit when running these manually in the GUI. I've updated it to 720p to make it a bit more standard.

sam-drumm (approved):

PR #452, opened 2024-10-28
Applicant under 18 test

sam-drumm:
https://hackney.atlassian.net/browse/TS-1763?atlOrigin=eyJpIjoiMWJjNjcyOTcwOTc1NGQ1MWIzMDY2NjNhYTZjYTMxOWMiLCJwIjoiaiJ9

**What**
- Local e2e test for applicant under 18,
- Refactors the confirmData into a try catch awaiting the two actions dispatched to get rid of the error below currently throwing on test. 

Screenshot 2024-10-28 at 09 49 47

### DIFF
--- a/cypress/e2e/local/ineligibleApplicantSucessfullyRejected.cy.ts
+++ b/cypress/e2e/local/ineligibleApplicantSucessfullyRejected.cy.ts
@@ -1,0 +1,118 @@
+import { faker } from '@faker-js/faker/locale/en_GB';
+import HomePage from '../../pages/home';
+import SignInPage from '../../pages/signIn';
+import VerifyPage from '../../pages/verify';
+import StartPage from '../../pages/start';
+import {
+  generateEmailAddress,
+  getRandomGender,
+  TitleEnum,
+} from '../../../testUtils/personHelper';
+import AgreeTermsPage from '../../pages/agreeTerms';
+import ApplyHouseholdPage from '../../pages/household';
+import ApplyExpectPage from '../../pages/apply/expect';
+import RejectionPage from '../../pages/rejection';
+
+import ApplyResidentSummaryPage from '../../pages/apply/[resident]/summary';
+
+// user is under 18
+const birthDate = faker.date.birthdate({ mode: 'age', min: 15, max: 17 });
+const title = faker.helpers.enumValue(TitleEnum);
+const phoneNumber = faker.phone.number();
+const mainApplicantFirstName = faker.person.firstName();
+const mainApplicantLastName = faker.person.lastName();
+const postcode = 'A1 2BC';
+const email = generateEmailAddress();
+const applicationId = faker.string.uuid();
+const verificationCode = faker.number
+  .int({ min: 100000, max: 999999 })
+  .toString();
+
+const fillInTheSignUpForm = () => {
+  StartPage.getTitleDropdown().select(title);
+  StartPage.getFirstNameInput().type(mainApplicantFirstName);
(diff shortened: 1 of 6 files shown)


pages/apply/[resident]/summary.tsx
```diff
@@ -81,31 +81,26 @@ const UserSummary = (): JSX.Element => {
     return <Custom404 />;
   }
 
-  const onConfirmData = () => {
+  const onConfirmData = async () => {
```
Comment (Duslerke) [inline pages/apply/[resident]/summary.tsx:84]: Should the button `await` the call now that it's async, or does that not matter in this particular context?

Comment (Duslerke) [reply pages/apply/[resident]/summary.tsx:84]: Oh, nvm.
I'm guessing it doesn't because... the button just triggers this background job.
And then if this job succeeds, only then the user gets redirected?

cypress/e2e/local/ineligibleApplicantSucessfullyRejected.cy.ts
```diff
@@ -0,0 +1,118 @@
...
+    it(`rejects an applicant under the age of 18`, () => {
+      cy.intercept(
+        {
+          method: 'GET',
+          path: '/api/address/*',
+        },
+        addressSearchAPIResponse
+      ).as('addressSearchMock');
+
+      cy.viewport(1000, 1000);
```
Duslerke, line 78: How come we limit this to 1000x1000?

cypress/e2e/local/ineligibleApplicantSucessfullyRejected.cy.ts
```diff
@@ -0,0 +1,118 @@
...
+        postcode: `${postcode}`,
+        UPRN: 11111111111,
+      },
+    ],
+    page_count: 1,
+    total_count: 1,
+  },
+};
+
+Cypress._.times(1, () => {
```
Duslerke, line 63: If this runs only once, perhaps, we don't need this wrapper?

cypress/e2e/local/ineligibleApplicantSucessfullyRejected.cy.ts
sam-drumm: default is 1000x600 this just make it easier to see when running in the UI.

cypress/e2e/local/ineligibleApplicantSucessfullyRejected.cy.ts
sam-drumm: It's mostly here because it's been useful as a local test to run multiple times to shake out any errors.

Duslerke: Is it still needed? :thinking:

Duslerke: If not, then it's better to remove as it would be comparable to dead code.

pages/apply/[resident]/summary.tsx
```diff
@@ -81,31 +81,26 @@ const UserSummary = (): JSX.Element => {
...
-        .unwrap()
-        .then(() => {
-          router.push('/apply/not-eligible');
-        })
-        .catch(() => {
-          setIsSaving(false);
-          setUserError('Unable to update application');
-          scrollToError();
-        });
+      try {
```
Comment (LBHTKarki) [inline pages/apply/[resident]/summary.tsx:92]: Should we add additional tests to ensure `sendDisqualifyEmail` call can fail without cancelling `disqualifyApplication`?

LBHTKarki (approved):

PR #454, opened 2024-10-28
TS-1761 applicant with medical condition

LBHTKarki:
Local e2e check that application is eligible when main applicant has a medical condition.


Duslerke (approved):
There's quite a lot of repetition between the tests of past few PRs.

Makes me wonder whether some cypress step groups could be extracted into shared functions with parameters? :thinking:

LBHTKarki:
> There's quite a lot of repetition between the tests of past few PRs.
> 
> Makes me wonder whether some cypress step groups could be extracted into shared functions with parameters? 🤔

Yes, there is bit of a repetition in these. I'll do a tidy up PR to extract some common elements to helper functions.

PR #462, opened 2024-10-30
Ts 1786 hr pipeline cleanup

sam-drumm:
https://hackney.atlassian.net/browse/TS-1786

**What**
- Reverts to testing Firefox in Pipeline after resolving both the video recording exclusion and the test failure, 
- Updates cypress to latest version,
- Uses mainApplicant to test age in useEffect as using currentResident triggers a 404,
- Uses index as key as applicant.person?.id was possibly undefined.

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -116,6 +116,7 @@ export default defineConfig({
     videoCompression: true,
   },
   component: {
+    experimentalJustInTimeCompile: true,
     devServer: {
       framework: 'next',
       bundler: 'webpack',
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -87,6 +87,8 @@ export default defineConfig({
 
       on(
         'after:spec',
+        // after the test has run, only save the video exists and if the test failed.
+        // https://docs.cypress.io/api/node-events/after-spec-api
         (spec: Cypress.Spec, results: CypressCommandLine.RunResult) => {
           if (results && results.video && results.stats.failures === 0) {
             if (existsSync(results.video)) unlinkSync(results.video);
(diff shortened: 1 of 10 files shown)


cypress/e2e/pages/apply/household/add-person.cy.ts
```diff
@@ -45,8 +44,7 @@ describe('Add person to household ', () => {
     HouseholdPage.visit();
     HouseholdPage.getHouseholdPage().should('be.visible');
 
-    AddPersonPage.visit();
-    AddPersonPage.getAddPersonPage().should('be.visible');
```
LBHTKarki, line 49: I guess this could be left as a check after clicking the add person button to ensure we are on the correct page

cypress.config.ts
```diff
@@ -83,18 +84,16 @@ export default defineConfig({
...
-              console.log('keeping the video %s', results.video);
-            } else {
-              unlinkSync(results.video);
-            }
+
+      on(
+        'after:spec',
+        (spec: Cypress.Spec, results: CypressCommandLine.RunResult) => {
+          if (results && results.video && results.stats.failures === 0) {
+            if (existsSync(results.video)) unlinkSync(results.video);
```
Duslerke, line 92: This could use some code comments for those not familiar with cypress configuration.

Duslerke (approved):
> "Reverts to testing Firefox in Pipeline after resolving both the video recording exclusion and the test failure."

So why exactly was Firefox failing when Chrome wasn't?

cypress.config.ts
sam-drumm: It was actually failing on this after:spec. I couldn't find a way to set the video:true to conditional on the browser as it's not set till after the config is initiated. 

So this was always causing issues when it tried to find the video file, but it didn't exist because Firefox fails at the record step. 

The existsSync step gets around that.

LBHTKarki (approved):

PR #467, opened 2024-10-31
Ts 1791 hr investigate safari test failures

sam-drumm:
https://hackney.atlassian.net/browse/TS-1791?atlOrigin=eyJpIjoiMDY1Yzk4MzAzNDRkNDI5ZWFjNGFiODk1Y2I3MGRmYzAiLCJwIjoiaiJ9

**What**
- Sets cypress tests to run on merge into main as well as development,
- Adds a type check that works on safari browser,

**Why**
- All test suites should run on a build from main,
- Tests are failing with Webkit - this adapts the error typing slightly so that all browsers show the correct error message.


martapederiva (approved):

lib/utils/errorHelper.ts
```diff
@@ -0,0 +1,5 @@
+export const checkError = (error: unknown) => {
```
LBHTKarki, line 1: To make it more clear what this is checking for would some other name be more descriptive? For example something like `isAssignableToError`?

sam-drumm: Yeah good point - I might just leave a comment as to why it exists also.

LBHTKarki (approved):

PR #469, opened 2024-11-04
TS-1793 improve sonar reliability scrore

LBHTKarki:
## WHAT
SonarCloud has flagged two issues that affect the app's reliability. This update addresses those issues.

## WHY
We want to aim for a reliability score 'A', so any issues that prevents us getting there should be fixed.

## HOW
1. Address _"Expected an assignment or function call and instead saw an expression"_ issue by removig the `.addDefaultcase` from the `hrApiCallsStatus` slice altogether. This slice has extra reducers only and the default state is set on init, so it's not necessary to have the defaultCase in place.
2. Address _"React Hook "useEffect" is called conditionally"_ issue by moving the 404 return after `useEffect`

### DIFF
--- a/pages/apply/[resident]/your-situation.tsx
+++ b/pages/apply/[resident]/your-situation.tsx
@@ -37,12 +37,6 @@ const YourSituation = (): JSX.Element => {
   );
   const saveApplicationStatus = useAppSelector(selectSaveApplicationStatus);
 
-  if (!applicant) {
-    return <Custom404 />;
-  }
-
-  const baseHref = `/apply/${applicant.person?.id}`;
-
   useEffect(() => {
     if (saveApplicationStatus?.callStatus === ApiCallStatusCode.FULFILLED) {
       setIsSaving(false);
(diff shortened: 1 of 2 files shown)


pages/apply/[resident]/your-situation.tsx
```diff
@@ -57,6 +51,12 @@ const YourSituation = (): JSX.Element => {
     }
   }, [saveApplicationStatus?.callStatus]);
 
+  if (!applicant) {
+    return <Custom404 />;
+  }
+
+  const baseHref = `/apply/${applicant.person?.id}`;
```
Comment (Duslerke) [inline pages/apply/[resident]/your-situation.tsx:58]: This should be above the useEffect as useEffect is using it.

Comment (LBHTKarki) [reply pages/apply/[resident]/your-situation.tsx]: That's true, I've address that in https://github.com/LBHackney-IT/lbh-housing-register/pull/469/commits/563e59b3d874a90cc4c69e8637c651cc9c442266. Test here https://github.com/LBHackney-IT/lbh-housing-register/blob/abce360397deaf0fa8dbe3148b9962206df0faf5/cypress/e2e/pages/apply/%5Bresident%5D/your-situation.cy.ts#L120 ensures the page doesn't throw an error when applicant is undefined on render.

Duslerke (approved):

PR #480, opened 2024-11-11
Ts 1810 housing register ethnicity questions

sam-drumm:
https://hackney.atlassian.net/browse/TS-1810?atlOrigin=eyJpIjoiOGY4NzA4OTVkZTQxNDkwMWExMjIxYjJhZWIyMWIzODQiLCJwIjoiaiJ9

**What**
User testing noted an error with the setting of ethnicity data.

Previously, the useApiCall status hook was pushing to the next page after data was saved. In the case of the ethnicity page, the second part of the form on the page was being skipped over after the initial save event. This change addresses this, so a user can set values in the second level of the form.

**Why**
Makes the hook more flexible to scenarios like this and ensures the user can engage with the whole form.

**How**
This changes the hook so if a path isn't provided, it won't push to a new page. In the case of this page, the path is only set on the final step of the ethnicity form.

### DIFF
--- a/cypress/e2e/local/eligibleApplicantWithMedicalCondition.cy.ts
+++ b/cypress/e2e/local/eligibleApplicantWithMedicalCondition.cy.ts
@@ -22,6 +22,7 @@ import {
   visitHomepageSignInAndVerify,
 } from '../../support/locale2eTestsHelper';
 import { interceptAddressSearchAPI } from '../../support/intercepts';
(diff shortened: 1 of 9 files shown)


LBHTKarki:
`managerActions` local e2e test gets stuck in the second ethnicity question. Would you be able to update that in this same PR?

LBHTKarki (approved):
👍

PR #482, opened 2024-11-11
Updates config for test before build

sam-drumm:
**What**
Ensures tests and sonar scan are complete before building to stage or development sites.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,10 +224,6 @@ workflows:
       - run-tests:
           requires:
             - install-dependencies
-      - sonar-scan:
-          context: SonarCloud
-          requires:
-            - install-dependencies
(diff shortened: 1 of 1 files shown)


LBHJHeppinstall (approved):

.circleci/config.yml
```diff
@@ -247,6 +247,8 @@ workflows:
       - assume-role-development:
           context: api-assume-role-housing-development-context
           requires:
+            - run-tests
+            - sonar-scan
```
Duslerke, line 251: sonar-scan step requires run-tests, as such here on line 250 requiring both the sonar scan, and its requirement run-tests is not needed.

Duslerke (approved):

LBHTKarki (approved):

PR #485, opened 2024-11-27
Speed up end to end tests in CircleCI 

spikeheap:
This PR introduces end to end tests to CircleCI builds on all branches.

Prior to this PR, end to end tests were only run on the `development` and `main` branches because they take around 11-15 minutes to complete. A typical full pipeline run including deployment looked like this:

Screenshot 2024-11-27 at 12 31 (Brave Browser)@2x

This PR makes a bunch of changes to the CircleCI config to improve clock time of builds:

1. Remove the separate dependency installation step, instead relying on the Node orb's built in caching of  `~/.npm` for efficient `npm ci` dependency installs.
2. Consolidate the   `npm build` step and cache the build artefact, preventing the deployment steps from having to rebuild the app. This improves overall deployment time without impacting feature branch test runs.
3. Consolidate the "assume AWS role" steps into the deployment job, removing the job spin up time. **Note this hasn't been tested because it runs on `development` and `main` branches only. This portion can be rolled back if the assumption that python is available on the  `aws-cli` executor isn't true.
4. Introduce parallel testing for Cypress. Using built-in CircleCI tooling and a script to clean up Cypress' test output we can split tests across multiple nodes, bringing the elapsed time down to just over 3 minutes from.
5. Bump the resource class from `Medium` to `Medium+` becuase the nodes were hitting max CPU.

There are also a minor tweak to the deploy command to use `npx` rather than an NPM install step. The effect is to apply the same command.

Following this change the CircleCI build looks like this for a feature branch:

Screenshot 2024-11-27 at 12 33 (Brave Browser)@2x

I haven't timed (or tested) the deployment job but I expect that to save around 1.5 minutes in spin-up and `npm build` time for each environment.

### Overall impact on feature branch builds

The build time including end to end tests is approximately the same as other branches without these changes applied:

Screenshot 2024-11-27 at 12 50 (Brave Browser)@2x
Screenshot 2024-11-27 at 12 52 (Brave Browser)@2x

### Considerations

1. This increases the cost of the CircleCI builds, both by running end to end tests for every branch and by parallelising the runs - around half of the parallel run is spent in setup. The resource class has been bumped as well. If this is problematic we can experiment with less parallelisation or fewer browser permutations for feature branches.
2. **The deployment pipelines haven't been tested (!)**. It would be reasonable to test this by deploying to development, either by just merging this PR and reverting it if it fails, or by changing the branch filter to allow a development deploy from the feature branch. I'm happy with either approach – what would work best for you?

### Future work

It _feels_ like there's still room for improvement around dependency installation and spin-up time, for example feature branches could save 20 seconds by not installing (or testing on) Firefox. If we could avoid the `npm run build` step for Cypress testing that could save around a minute too. 

I didn't pursue these because my goal was just to enable end to end testing in feature branches.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -159,6 +159,10 @@ jobs:
     executor: aws-cli/default
     steps:
       - checkout
+      - aws_assume_role/assume_role:
+          account: $AWS_ACCOUNT_DEVELOPMENT
+          profile_name: default
+          role: 'LBH_Circle_CI_Deployment_Role'
       - node/install-packages
       - run:
           name: deploy
@@ -168,6 +172,10 @@ jobs:
     executor: aws-cli/default
     steps:
       - checkout
+      - aws_assume_role/assume_role:
+          account: $AWS_ACCOUNT_STAGING
+          profile_name: default
+          role: 'LBH_Circle_CI_Deployment_Role'
       - node/install-packages
       - run:
           name: deploy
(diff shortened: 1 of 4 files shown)


LBHTKarki (approved):
This should tidy and speed things up quite a bit 👍

PR #486, opened 2024-12-06
Updates the analytics config

sam-drumm:
https://hackney.atlassian.net/browse/TS-1849?atlOrigin=eyJpIjoiZjY2YzhmNmRjNTNmNGNmOWFlMTllZjQ5NTc4MDYwMGMiLCJwIjoiaiJ9

**What**
- GTM ID is in the head but not being received. 
- This PR removes the ID from _App and uses the recommended configuration in __document only.

### DIFF
--- a/pages/_document.tsx
+++ b/pages/_document.tsx
@@ -1,4 +1,5 @@
 import Document, { Html, Head, Main, NextScript } from 'next/document';
+import Script from 'next/script';
 
 import { ReactElement } from 'react';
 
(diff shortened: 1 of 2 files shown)


pages/_document.tsx
```diff
@@ -1,20 +1,32 @@
...
-        <body className="govuk-template__body lbh-template__body js-enabled">
-          <Main />
-          <NextScript />
-          <noscript
+        <Head>
+          <Script
+            id="gtm"
             dangerouslySetInnerHTML={{
-              __html: `<iframe src="https://www.googletagmanager.com/ns.html?id=${process.env.NEXT_PUBLIC_GTM_ID}" height="0" width="0" style="display: none; visibility: hidden;" />`,
+              __html: `(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({"gtm.start":new Date().getTime(),event:"gtm.js"});var f=d.getElementsByTagName(s)[0],j=d.createElement(s),dl=l!="dataLayer"?"&l="+l:"";j.async=true;j.src="https://www.googletagmanager.com/gtm.js?id="+i+dl;f.parentNode.insertBefore(j,f);})(window,document,"script","dataLayer","${process.env.NEXT_PUBLIC_GTM_ID}");`,
```
Duslerke, line 14: I don't get it.
What does this line14 do that the block lines17-29 doesn't? :thinking:

Duslerke: Clearly these have differences, but this `__html` and the string's that's fed into it roles are unclear.

sam-drumm: So my understanding is the iframe is a fallback if JS is disabled.

Duslerke (approved):

pages/_document.tsx
Duslerke: That makes sense.
I was thinking that maybe this was duplicated by accident :sweat_smile:

PR #525, opened 2026-04-02
Feature/upgrade next

sam-drumm:
## What

Upgrades the Housing Register to Next.js 16, React 19, and current Node (24), TypeScript, ESLint 9, Jest, Cypress, and Sentry versions.

## How

**Runtime / framework**
- Next 16, React 19, next-redux-wrapper 8 (major updates for all three).
- All other libraries pushed to the latest supported version.
- `lodash.throttle` replaced with full `lodash` package because of importing errors; `csv-writer` removed (was unused, probably leftover from a novalet feature that was never implemented).

**Tooling**
- Node >=24 enforced via `.nvmrc` and `package.json` `engines`; Lambda runtime updated to `nodejs24.x`.
- ESLint 9 flat config `eslint.config.mjs`; legacy `.eslintrc` removed.
- Jest config moved to `jest.config.js`; Babel presets updated for React 19 / TS 5.
- TypeScript / `@types/*` updated throughout for React 19 compatibility.

**Sentry**
- `@sentry/nextjs` v10: `instrumentation.ts` + `instrumentation-client.ts` replace the old
  `sentry.client.config.js`; `sentry.server.config.js` and `sentry.edge.config.js` updated
  for v10 API.
- API handlers wrapped with `wrapApiHandlerWithSentry`.

**UI**
- Dialog component migrated from `lbh-frontend` modal to `@radix-ui/react-dialog`. `@reach/dialog` was pulled in from `lbh-frontend` and was archived in 2023 and not supported for React 19.
- Link typing updated as Link now renders the <a> itself.

**E2E**
- In-process HTTP mocking via `E2E_HTTP_MOCKS`: `pages/api/e2e/nock`,
  `pages/api/e2e/clear-nock`, `pages/api/e2e/generate-token`, and
  `lib/server/e2eHttpMocks.ts`. Spinning up nock within the Cypress start process no longer possible.
- Cypress commands and specs updated to use mocked APIs; `cypress:open:local` and
  `e2e:run:local` scripts added for `LOCAL_E2E` mode.

**Bug fixes**
- Apply resident: `router.isReady` / loading guard added before treating a missing
  resident as a 404; verify page redirect ordering fixed.
- API `auth/generate`: safer control flow (removed invalid `break` in `catch` as flagged by Sonar);
  E2E-only logging gated on `E2E_HTTP_MOCKS` so Jest no longer spams `console.error`.
- `generate.spec.ts`: handler now properly `await`ed; brittle global `JSON.parse`
  call-count assertion on Sentry-wrapped 500 path removed.

**Lambda deployment**
- `output: standalone` + `distDir: build/_next` in `next.config.js`.
- `scripts/prepare-lambda-standalone.cjs`: installs `serverless-http` into the
  standalone bundle, seals symlinks (prevents Serverless zip bloat), prunes SWC
  binaries / non-linux esbuild / source maps / docs to stay under the 250 MB
  unzipped limit.
- `lambda.js` rewritten: extracts the embedded Next.js config from `standalone/server.js`. `/_next/static/**` assets are served directly from the Lambda filesystem with explicit `Content-Type` headers to avoid MIME-type mismatches that `serverless-http` corrupted.
- `serverless.yml`: packaging updated to include `build/_next/standalone/**` +
  `build/_next/static/**` (source maps excluded). Environment variables moved from
  inline `${ssm:...}` lookups to `${param:...}`.

**CI**
- CircleCI executor updated to `cimg/node:24.14-browsers`.
- Cypress job: `E2E_HTTP_MOCKS`, mock API URLs, and `AUTHORISED_ADMIN_GROUP` set at
  job level.

## Notes

`serverless.yml` reads environment variables as Serverless Framework parameters (`${param:...}`) from the central parameters stores rather than directly from SSM. The deploy command supply these via `--param` flags.

Non-secret config values are now consumed from the central parameter store, and secrets are still consumed from the account parameters. These values have been updated to secret strings where previously stored as strings.


lib/server/e2eHttpMocks.ts
```diff
@@ -0,0 +1,69 @@
...
+    typeof (scope as unknown as Record<string, unknown>)[verb] !== 'function'
+  ) {
+    throw new Error(`E2E nock: unsupported method "${method}"`);
+  }
+
+  const interceptor = (
+    scope as unknown as {
+      [k: string]: (p: string) => nock.Interceptor;
+    }
+  )[verb](path);
```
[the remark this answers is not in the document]
sam-drumm: This is on mock paths so no risk

components/content/CookieBanner.tsx
```diff
@@ -6,14 +6,25 @@ const CookieBanner: React.FC = (): JSX.Element => {
...
-      }
+      if (typeof window === 'undefined') return;
+
+      const { CookieBanner: LbhCookieBanner } = await import('lbh-frontend');
+      if (cancelled) return;
+
+      const root = ref.current;
+      if (!root) return;
+
+      new LbhCookieBanner(root).init();
```
sam-drumm, line 20: fixes intermittent pipeline failure on apply 404 under React Strict Mode

.circleci/config.yml
```diff
@@ -72,6 +81,141 @@ exclude_combinations:
     width: 320
 
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Commands ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ #
+commands:
+  install-python3:
+    description: 'pip install for aws-cli'
+    steps:
+      - run:
+          name: Install python3.12-venv
```
LBHTKarki: Would it be worth bumping this up to latest i.e. 3.14?

components/admin/MainApplicantForm.tsx
```diff
@@ -138,6 +141,13 @@ export default function MainApplicantForm({
           >
             {({ touched, isSubmitting, errors, isValid }) => {
               const isTouched = Object.keys(touched).length !== 0;
+              const formatErrorTitle = (
```
LBHTKarki: would it be worth moving this to a helper function?

.circleci/config.yml
sam-drumm: Actually this breaks the pipeline. So sticking with .12 here.

LBHTKarki (approved):

PR #530, opened 2026-04-27
Test/fix local tests

sam-drumm:
What
-- 
The previous flow was flaky on first load due to page changes where similarly named ids existed across pages. This PR fixes the local cypress submit flow around `/apply/overview` → `/apply/submit/additional-questions` → `/apply/submit/ethnicity-questions` by using page-specific test IDs.

This also adds a small visual feature that was requested to display the user names of those with stated medical needs. As user had to open each application to find the medical needs, this was a small FE addition to save time.

Screenshot 2026-05-01 at 15 17 26

How
--
- Uses page objects instead of flaky `.lbh-button`/radio selectors.
- Added test IDs for:
  - overview “Save and continue”
  - additional-questions submit button
- Moved ethnicity radio interactions into existing `EthnicityPage` helpers.
- Removed incorrect summary “save and continue” page-objects.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -40,8 +40,5 @@ NOTIFY_TEMPLATE_MEDICAL_NEED=
 # Activity History API
 ACTIVITY_HISTORY_API=http://localhost:3600/api/v1
 
-# Set to true to exclude running e2e/local test folder in the environment. These tests require a running localstack config outlined in readme.
-LOCAL_E2E=
-
 # Enable /api/e2e/* routes for Cypress (Nock + JWT helpers in the Next.js process). Required for mocked e2e.
 E2E_HTTP_MOCKS=
--- a/__tests__/pages/api/auth/generate.spec.ts
+++ b/__tests__/pages/api/auth/generate.spec.ts
@@ -29,7 +29,7 @@ const mockCreateAuthRequest: CreateAuthRequest = {
   email: email,
 };
 
-/** String body so the handler takes the `JSON.parse(req.body)` branch (see generate.tsx). */
+/** String body so the handler takes the `JSON.parse(req.body)` path. */
 const mockCreateAuthRequestBody = JSON.stringify(mockCreateAuthRequest);
 
 const mockCreateAuthResponse: CreateAuthResponse = { success: true };
--- a/__tests__/pages/api/auth/generate.spec.ts
+++ b/__tests__/pages/api/auth/generate.spec.ts
@@ -29,6 +29,9 @@ const mockCreateAuthRequest: CreateAuthRequest = {
   email: email,
 };
 
+/** String body so the handler takes the `JSON.parse(req.body)` branch (see generate.tsx). */
+const mockCreateAuthRequestBody = JSON.stringify(mockCreateAuthRequest);
+
 const mockCreateAuthResponse: CreateAuthResponse = { success: true };
 
 jest.mock('../../../../lib/gateways/applications-api', () => ({
@@ -47,7 +50,8 @@ describe('POST', () => {
 
   const requestOptions: RequestOptions = {
     method: 'POST',
-    body: mockCreateAuthRequest,
+    // createMocks accepts a JSON string body; `Body` typings omit `string`.
+    body: mockCreateAuthRequestBody as unknown as RequestOptions['body'],
   };
 
   it('returns status code 200 when JSON.parse succeeds and verify code is created successfully', async () => {
@@ -69,7 +73,7 @@ describe('POST', () => {
 
   it('returns status code 400 when JSON.parse fails', async () => {
     jsonParseSpy.mockImplementationOnce(() => {
-      throw SyntaxError('Unexpected token');
+      throw new SyntaxError('Unexpected token');
     });
 
     createVerifyCodeMock.mockReturnValueOnce(mockCreateAuthResponse);
@@ -97,7 +101,7 @@ describe('POST', () => {
 
     const reqOptions: RequestOptions = {
       method: 'POST',
-      body: mockCreateAuthRequest,
+      body: mockCreateAuthRequestBody as unknown as RequestOptions['body'],
     };
 
     const { req, res }: { req: ApiRequest; res: ApiResponse } =
--- a/cypress/pages/additionalQuestions.ts
+++ b/cypress/pages/additionalQuestions.ts
@@ -6,6 +6,26 @@ class AdditionalQuestionsPage {
   static visit() {
     cy.visit(`/apply/submit/additional-questions`);
   }
+  static getSaveAndContinueButton() {
+    const testId = 'test-additional-questions-save-and-continue';
+    return AdditionalQuestionsPage.getAdditionalQuestionsPage().find(
+      `[data-testid="${testId}"]`,
+    );
+  }
+  static clickSaveAndContinue() {
+    return AdditionalQuestionsPage.getSaveAndContinueButton()
+      .scrollIntoView()
+      .should('be.visible')
+      .and('not.be.disabled')
+      .click();
+  }
+  static expectLoaded() {
+    cy.url({ timeout: 30000 }).should(
+      'include',
+      '/apply/submit/additional-questions',
+    );
+    AdditionalQuestionsPage.getAdditionalQuestionsPage().should('exist');
+  }
   static getAdditionalQuestionsNotes() {
     const testId = 'test-checkbox-conditional-container';
     return cy.get(`[data-testid*="${testId}"]`);
--- a/cypress/pages/additionalQuestions.ts
+++ b/cypress/pages/additionalQuestions.ts
@@ -7,7 +7,7 @@ class AdditionalQuestionsPage {
     cy.visit(`/apply/submit/additional-questions`);
   }
   static getSaveAndContinueButton() {
-    const testId = 'test-additional-questions-save-and-continue';
+    const testId = 'test-submit-form-button';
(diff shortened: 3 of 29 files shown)


.env.sample
```diff
@@ -40,8 +40,5 @@ NOTIFY_TEMPLATE_MEDICAL_NEED=
 # Activity History API
 ACTIVITY_HISTORY_API=http://localhost:3600/api/v1
 
-# Set to true to exclude running e2e/local test folder in the environment. These tests require a running localstack config outlined in readme.
```
sam-drumm, line 43: Set when cypress local commands are run

pages/api/auth/verify.tsx
```diff
@@ -1,17 +1,40 @@
 import { wrapApiHandlerWithSentry } from '@sentry/nextjs';
+import axios from 'axios';
 import { StatusCodes } from 'http-status-codes';
 import type { NextApiHandler, NextApiRequest, NextApiResponse } from 'next';
+import type { VerifyAuthRequest } from '../../../domain/HousingApi';
 import { confirmVerifyCode } from '../../../lib/gateways/applications-api';
 import { setAuthCookie } from '../../../lib/utils/users';
 
+function parseVerifyBody(req: NextApiRequest): VerifyAuthRequest | null {
```
sam-drumm, line 9: Added to consistently sanitise the auth request and provide a 400 when invalid.

pages/api/auth/generate.tsx
```diff
@@ -41,7 +45,15 @@ const endpoint: NextApiHandler = async (
...
-    const data = await createVerifyCode(request);
+    // In local Cypress we pre-seed a resident token cookie with application_id.
+    // Forward it so API can create/verify against the same application record.
+    const cookieApplicationId = getUser(req)?.application_id;
+    const requestWithApplicationId: CreateAuthRequest = {
+      ...request,
+      applicationId: request.applicationId ?? cookieApplicationId,
+    };
+
+    const data = await createVerifyCode(requestWithApplicationId);
```
sam-drumm, line 57: For local cypress the try uses getUser.application_id, and reads housing_user from a JWT cookie. Normally that JWT is validated with jsonwebtoken.verify(...), so the application_id is trusted. We use request.applicationId ?? cookieApplicationId, so an explicit applicationId in the request always takes priority in normal prod mode.

So when we set SKIP_VERIFY_TOKEN === 'true', getUser uses jsonwebtoken.decode(...).

pages/api/applications/index.ts
```diff
@@ -31,6 +32,12 @@ const endpoint: NextApiHandler = async (
             .json({ message: 'Unable to get application' });
         }
       } catch (error) {
+        if (axios.isAxiosError(error) && error.response?.status) {
+          res
+            .status(error.response.status)
+            .json({ message: 'Unable to get application' });
```
sam-drumm, line 38: Just going to leave this - was added during testing but still has some value.

lib/store/auth.ts
```diff
@@ -43,9 +44,11 @@ export const confirmVerifyCode = createAsyncThunk(
 
     if (res.ok) {
       return (await res.json()) as VerifyAuthResponse;
-    } else {
-      return rejectWithValue(`Unable to confirm verify code (${res.status})`);
     }
+    if (res.status === 404) {
+      return rejectWithValue(Errors.VERIFY_CODE_NOT_CONFIRMED);
```
sam-drumm, line 49: Again was added during debugging but still has some value

cypress/support/intercepts.ts
```diff
@@ -1,5 +1,24 @@
...
+    id: '00000000-0000-4000-8000-000000000001',
+    createdAt: new Date().toISOString(),
+  },
+];
+
+/**
+ * Stubs the Next BFF evidence endpoint from the browser. Use for `LOCAL_E2E` when the
+ * Housing Register API has no Evidence service (e.g. localstack) — no `E2E_HTTP_MOCKS` required.
+ */
+export const interceptLocalE2eApplicationEvidencePost = () => {
```
sam-drumm, line 15: If I don't do this in local mode, I can never complete an application. Not sure how/why it worked before but this stubs the evidence api failure (api not in localstack) to allow application to complete.

__tests__/pages/api/auth/generate.spec.ts
```diff
@@ -97,7 +101,7 @@ describe('POST', () => {
 
     const reqOptions: RequestOptions = {
       method: 'POST',
-      body: mockCreateAuthRequest,
+      body: mockCreateAuthRequestBody as unknown as RequestOptions['body'],
```
sam-drumm, line 104: mockCreateAuthRequest (object) parse never ran. That made tests like JSON.parse fails flaky, because there was nothing to do

cypress/pages/apply/overview.ts
```diff
@@ -18,6 +18,17 @@ class ApplyOverviewPage {
       .should('be.visible')
       .click({ force: true });
   }
+
+  static getSaveAndContinueToAdditionalQuestionsLink() {
+    return cy.get('[data-testid="test-apply-overview-save-and-continue"]');
```
Duslerke, line 23: having `test` within a value of "data-**test**id feels redundant, but for the sake of this PR, it can be kept as is.

cypress/pages/additionalQuestions.ts
```diff
@@ -6,6 +6,26 @@ class AdditionalQuestionsPage {
...
+  }
+  static clickSaveAndContinue() {
+    return AdditionalQuestionsPage.getSaveAndContinueButton()
+      .scrollIntoView()
+      .should('be.visible')
+      .and('not.be.disabled')
+      .click();
+  }
+  static expectLoaded() {
+    cy.url({ timeout: 30000 }).should(
```
Duslerke, line 23: I'm quite curious. Why is such a long timeout needed to get the URL?

My understanding could be wrong, but the URL state does not depend on page load unless we're dealing with redirects.

pages/api/applications/index.ts
Duslerke: The `NextApiHandler` is getting this Sonar scan error:

Image

pages/api/applications/index.ts
```diff
@@ -31,6 +32,12 @@ const endpoint: NextApiHandler = async (
             .json({ message: 'Unable to get application' });
         }
       } catch (error) {
+        if (axios.isAxiosError(error) && error.response?.status) {
```
Duslerke, line 35: What is the role of `&& error.response?.status`?

pages/api/auth/generate.tsx
```diff
@@ -27,7 +28,10 @@ const endpoint: NextApiHandler = async (
 
   let request: CreateAuthRequest;
   try {
-    request = JSON.parse(req.body) as CreateAuthRequest;
+    request =
+      typeof req.body === 'string'
+        ? (JSON.parse(req.body) as CreateAuthRequest)
+        : (req.body as CreateAuthRequest);
```
Duslerke, line 35: I think a doc-purpose code comment is needed here to explain why and when the `req.body` becomes `string`.

pages/api/auth/generate.tsx
Duslerke, line 57: I don't follow this logic... likely due to not having worked on this App before.
Might be a good idea to have @LBHTKarki 's eyes on this to make sure this is secure.

pages/api/auth/verify.tsx
```diff
@@ -1,17 +1,40 @@
...
+  try {
+    if (typeof req.body === 'string') {
+      return JSON.parse(req.body) as VerifyAuthRequest;
+    }
+    if (req.body && typeof req.body === 'object') {
```
Duslerke: It's not an `if`, but an `else if` as the when one is true, the other is guaranteed to be false.
Saves a tiny bit of compute.

.env.sample
```diff
@@ -40,8 +40,5 @@ NOTIFY_TEMPLATE_MEDICAL_NEED=
...
-LOCAL_E2E=
```
Duslerke, line 44: Why is this being removed from `.env.sample` if we have this line in `cypress.config.ts`:
```
LOCAL_E2E: process.env.LOCAL_E2E?.trim() === 'true' ? 'true' : undefined,
```
?

Seems to me like this env var is still being relied on for whatever logic.

sam-drumm: I could leave it in and yes it's in use but being set directly in the command. 

`"cypress:open:local": "LOCAL_E2E=true E2E_HTTP_MOCKS=true cypress open --browser chrome",` 

Just I think a bit easier to follow that way

cypress/pages/additionalQuestions.ts
sam-drumm: It's just a flake work around from the previous test which I could get rid of now I look at it again

sam-drumm: I might actually just leave this in. It works without it now but at least it's a small safety net on the '/apply/submit/additional-questions', and '/apply/submit/ethnicity-questions', paths.

cypress/pages/apply/overview.ts
sam-drumm: I think it's just the established pattern that was followed

pages/api/applications/index.ts
sam-drumm: To make sure it was an actual api failure with response code otherwise fall to generic

Duslerke (approved):

LBHTKarki (approved):

PR #532, opened 2026-05-12
build: clean opennext config

sam-drumm:
What
--

The lambda.js and prebuild/preparation/prune files to meet lambda limits were very hard to follow, and also possibly convoluted. While they 'worked' I want to try a cleaner solution that might also benefit other modern Next apps.

This switches from handcrafted lambda.js to using [OpenNext](https://opennext.js.org/) to handle the Next build into a Lambda ready bundle out of the box. 

How
--
* open-next.config.ts matches the same deployment - s3 hosted statick assets served via cloudfront 
* serverless user the prebuilt artifact, creates a bucket for public assets and adds a invalidation step for cloudfront
* package.json removes postbuild, standalone and serveless-http dependancies
* circle now runs build:opennext

Also
--
* Bumps Next to address security warnings,
* Fixes linting issues on api/reports/* where any was no longer being passed

Testing
--

* Current build on the development site.


.circleci/config.yml
```diff
@@ -216,6 +217,27 @@ commands:
               --param="evidenceStoreUrl=${NEXT_PUBLIC_EVIDENCE_STORE}" \
               --param="gtmId=${NEXT_PUBLIC_GTM_ID}" \
               --conceal
+      - run:
+          name: Sync static assets to S3 + invalidate CloudFront
```
Duslerke, line 136: Why do these need to be deployed separately?

.circleci/config.yml
```diff
@@ -252,13 +274,24 @@ jobs:
...
-          name: Build Next application
-          command: npm run build
+          name: Build Next.js + OpenNext bundle
+          command: npm run build:opennext
+      - run:
+          # Lambda expects `index.handler` at the zip root, so we zip from
+          # *inside* the directory rather than including the path prefix.
+          name: Package Lambda artifact
+          command: |
+            set -e
```
Duslerke: `set -e` Isn't this the default behaviour of CircleCI run step?

.circleci/config.yml
```diff
@@ -252,13 +274,24 @@ jobs:
...
+            mkdir -p build
```
Duslerke: Don't need `-p` when you're create a dir within current dir.

.circleci/config.yml
```diff
@@ -252,13 +274,24 @@ jobs:
...
+            (cd .open-next/server-functions/default && zip -qr ../../../build/opennext-server.zip .)
```
Duslerke: Would be more readable to split into 2 separate commands.
The `&&` has no effect anyway because the entire script block will fail anyway when the `cd` tries to change into non-existent dir.

.circleci/config.yml
```diff
@@ -252,13 +274,24 @@ jobs:
...
+            ls -lh build/opennext-server.zip
```
Duslerke: Can avoid printing user and perms:
```suggestion
            ls -sh build/opennext-server.zip
```

.circleci/config.yml
```diff
@@ -252,13 +274,24 @@ jobs:
...
       - persist_to_workspace:
           root: *workspace_root
           paths:
-            - build
-            - node_modules
+            - .next # used by run-cypress-e2e (next start)
+            - .open-next/assets # used by deploy (aws s3 sync)
+            - build/opennext-server.zip # used by deploy (serverless artifact)
+            - node_modules # used by run-cypress-e2e (next start)
```
Duslerke, line 209: Nice documentation

open-next.config.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Notes:
+ *  - OpenNext v4 always emits image-opt, warmer, and revalidation bundles —
+ *    there is no toggle to skip them. We simply don't deploy them.
+ *  - We override `imageOptimization.install` so OpenNext's
+ *    `installDependencies` short-circuits and skips the `npm install sharp`
+ *    step that runs by default.
+ *  - `dangerous.disableIncrementalCache` and `disableTagCache` strip the S3 /
+ *    DynamoDB cache wiring from the server bundle. This app has no ISR /
+ *    revalidateTag usage, so neither is needed and skipping them avoids
+ *    requiring the cache bucket + DDB table at deploy time.
```
Duslerke: What is it the cache of? I don't quite follow how DynamoDB ties into this as well.

serverless.yml
```diff
@@ -1,3 +1,29 @@
+# =============================================================================
+# OpenNext deployment for the Housing Register frontend.
+#
+# Build prerequisites (run before `serverless deploy --config serverless.yml`):
+#
+#   1. npm run build:opennext                # produces .open-next/
+#   2. mkdir -p build
+#   3. (cd .open-next/server-functions/default && zip -qr ../../../build/opennext-server.zip .)
+#   4. aws s3 sync .open-next/assets s3://<bucket-from-this-stack>/_assets \
+#        --cache-control "public,max-age=31536000,immutable"
```
Duslerke: I already saw this code within CircleCI pipeline.
I don't quite follow why it's listed here in the comments.

Duslerke: Ok, got it.
Might be better to just give just tell to look for a step with specific name within CCI config to avoid code duplication.

serverless.yml
```diff
@@ -1,3 +1,29 @@
...
+# OpenNext produces from the Next server adapter).
+#
+# Notes:
+#   - `package.artifact` points at a pre-built zip rather than using
+#     `package.patterns`, because the OpenNext output expects to be unpacked
+#     at the zip root (`index.mjs` next to `node_modules/`), and sls
+#     `patterns` can't easily strip a directory prefix.
+#   - Static assets are not packaged with the function — they live in S3 and
+#     are routed via CloudFront (two origins: API Gateway for SSR, S3 for
+#     `_next/static/*`, `BUILD_ID`, `favicon.ico`, `images/*`).
```
Duslerke, line 18: Is this just a thing that framework expects?

serverless.yml
```diff
@@ -59,6 +74,77 @@ functions:
...
+    #   /_assets/_next/static/...
+    #   /_assets/favicon.ico
+    #   /_assets/images/...           (formerly from public/images)
+    # CloudFront reaches it via an Origin Access Control (OAC)
+    # -------------------------------------------------------------------------
+    AssetsBucket:
+      Type: AWS::S3::Bucket
+      # Keep the bucket around if the stack is deleted or a failed update
+      DeletionPolicy: Retain
+      UpdateReplacePolicy: Retain
```
Duslerke: From what I understand, this isn't valuable data... just things that get added via deployment pipeline. So shouldn't need to 'Retain'.

serverless.yml
```diff
@@ -59,6 +74,77 @@ functions:
...
+          Rules:
+            - ObjectOwnership: BucketOwnerEnforced
+        LifecycleConfiguration:
+          Rules:
+            # This rule expires the accumulated old chunks after 30 days.
+            # This is long enough for any active session to complete.
+            - Id: expire-old-next-static-chunks
+              Status: Enabled
+              Prefix: '_assets/_next/static/'
+              ExpirationInDays: 30
```
Duslerke, line 102: What session are we talking about here?

My understanding is that this bucket holds build artefacts used by application, so there shouldn't be anything special here that would need to be held onto.

Additionally, doesn't the deployment script clear old files anyway?
I haven't done s3 sync before, but from my exp, sync typically removes files that don't exist in source, and uploads those that don't exist on target.

serverless.yml
```diff
@@ -84,19 +173,103 @@ resources:
...
+              AllowedMethods: [GET, HEAD]
+              CachedMethods: [GET, HEAD]
+              DefaultTTL: 86400
+              MaxTTL: 86400
+              MinTTL: 86400
+              ForwardedValues:
+                QueryString: false
+                Cookies:
+                  Forward: none
+              Compress: true
```
Duslerke, line 226: Are all these values actually required?
That's a lot of config, and I'm wondering... can't most of these keys use the defaults? Or are these values deviations from default that are absolutely necessary? In which case, it would be good to document why.

.circleci/config.yml
sam-drumm: Not sure what you mean here - do you mean why not get serverless to handle the asset upload in the previous step?

.circleci/config.yml
sam-drumm: yup - copy paste not needed.

open-next.config.ts
sam-drumm: I'll get rid of some of this. Next has some features (ISR) where you can get a page to regenerate on a set time that requires some additional DB setup. Not relevant to us

serverless.yml
sam-drumm: Pretty much seems to be the pattern that OpenNext uses. https://opennext.js.org/aws/inner_workings/architecture

serverless.yml
sam-drumm: I think the theory here is CF can't delete a bucket unless its empty. If I set it to delete it risks the stack being stuck in in a failed state, this is just a work around.

serverless.yml
sam-drumm: Session being a user who has 30 days to reload until navigation would break on the new app. 

the s3 sync only adds and updates files unless you specific delete which we're not doing.

serverless.yml
sam-drumm: Have cut where I can and put some more comments

package.json
```diff
@@ -58,14 +57,13 @@
     "lbh-frontend": "^3.6.1",
     "lodash": "^4.17.21",
     "nested-object-diff": "^1.1.0",
-    "next": "^16.2.3",
+    "next": "^16.2.6",
```
sam-drumm, line 60: we're already on 16.2.6 in the lock but this sets

serverless.yml
Duslerke: Got it, yeah you're right I forgot about this tiny quirk. You need clear the buckets and their versions manually.
Though I'm surprised serverless didn't provide an option you can use to explicitly tell it to clear data as well.
You'd think in clearing up unused resources, including S3 buckets like those where front-end source files live, etc... it would be common use case.

Duslerke: Nvm, I see you need a plugin for that.

.circleci/config.yml
Duslerke: nvm, answer is OpenNext expected architecture pattern.

CHANGELOG.md
```diff
@@ -2,58 +2,54 @@
...
+- sentry flags addresses ([16b2d09](https://github.com/LBHackney-IT/lbh-housing-register/commit/16b2d09c51657a18dfefae8f4406a1e793433d39))
+- set content type in headers ([266f55a](https://github.com/LBHackney-IT/lbh-housing-register/commit/266f55adf5a25460f3c9f5bb26e79b7955a93637))
+- set envs in cypress ([158a8b7](https://github.com/LBHackney-IT/lbh-housing-register/commit/158a8b7756d68bc2de05116a05b2a7c09ae2c28b))
+- sonar flag addressed ([6ea33d2](https://github.com/LBHackney-IT/lbh-housing-register/commit/6ea33d2b9a2579b4dafcf45b9ef675ef74bb0a12))
+- sonar flags ([5da4c39](https://github.com/LBHackney-IT/lbh-housing-register/commit/5da4c3993edb04a6134d2951673a81f31d3f6938))
+- sonar flags addressed ([f6d9c1f](https://github.com/LBHackney-IT/lbh-housing-register/commit/f6d9c1fd1410b0003f7949b96f9e34a722e903cf))
+- strip environment from paths ([4737837](https://github.com/LBHackney-IT/lbh-housing-register/commit/4737837d6a4077fdf7209141259d812684f18643))
+- test on branch ([59fccc8](https://github.com/LBHackney-IT/lbh-housing-register/commit/59fccc89160c761c7088f7679648047d9036fbaa))
```
Duslerke, line 35: Not the most useful CHANGELOG entries.
But I'm guessing these were generated via some tool?

Duslerke (dismissed):
Looks good. Hard to reason about some config not having tried this, or seen it work.
But let's give it a go and see how it works.

Miles-Alford (approved):

PR #535, opened 2026-05-18
Bugfix/broken auth

sam-drumm:
What
--
Fixes a production bug introduced in v2.2.0, a bug in axiosClients.ts, two linting errors in the report handlers, Next version bump and adjustments to the deployment workflow.

Root cause
---
v2.2.0 added two things to pages/api/auth/generate.tsx to support the local Cypress E2E test suite:

- applicationId forwarding — the local tests run against the local backend with a pre-seeded application. When a test user signs in, the handler reads the cookie (which contains application_id), and forwards it to the backend alongside the email. The backend connects the generated code to the correct application record.

- Body parsing — to make the above work, the handler needed to accept the request as either a string or a parsed object. When the test engages with the sign-in form, Next parses the body into an object before the handler sees it, so the original JSON.parse was parsing a plain object, producing [object Object].

Accepting an object body without then validating email then caused an issue. Previously JSON.parse({}) throwing an error was having a positive side effect — any request with a missing or empty body got a 400 before reaching the backend. After the change, an empty body ({}) was accepted and forwarded to the backend with email:null. This results in a 500 from the backend.

The "send a new code" button on /apply/verify triggered this in prod. The button dispatches createVerifyCode(email) where email = router.query.email as string. The same gap existed in pages/api/auth/verify.tsx.

Changes
---
pages/api/auth/generate.tsx Added guard after body parsing. If request.email is not correct, the handler returns 400 "email is required" before calling the backend. 

pages/api/auth/verify.tsx Same pattern applied to both email and code.

pages/apply/verify.tsx resendCode now returns early with a FE error if email is false, preventing any call.
The "Send a new code" button is disabled if email is unset.

lib/utils/axiosClients.ts Fixed a bug in where the auth header was set to "Bearer undefined" when no hackneyToken cookie was present.

pages/api/reports/internal/download.tsx and pages/api/reports/novalet/download/[fileName].tsx Fixed pre-existing TS errors flagging in pre-commit hook: AxiosHeaderValue includes boolean and AxiosHeaders in addition to string, which are incompatible with res.setHeader.

Also
---
* Production workflow has been updated. On merge to main, staging now builds. Release please builds the release PR in the background. On approval of this PR the production application is deployed.
* Grabs central parameter store directly from serverless - previously the bash command exporting the params didn't handle whitespace.
* Updates sonar badges in readme as the repo is now private
* Addresses security issues with Next (bump next version)
* Updates readme to more clearly explain the release workflow


pages/api/reports/internal/download.tsx
```diff
@@ -53,10 +53,20 @@ const endpoint: NextApiHandler = async (
...
-            res.setHeader('Content-Type', fileResponse.headers['content-type']);
+            const contentType = fileResponse.headers['content-type'];
+            res.setHeader(
+              'Content-Type',
+              typeof contentType === 'string'
+                ? contentType
+                : 'application/octet-stream',
+            );
+            const contentDisposition =
+              fileResponse.headers['content-disposition'];
```
sam-drumm, line 64: Just handling linting error 

```Argument of type 'AxiosHeaderValue | undefined' is not assignable to parameter of type 'string | number | readonly string[]'.
  Type 'undefined' is not assignable to type 'string | number | readonly string[]'.

File: pages/api/reports/novalet/download/[fileName].tsx [hunk 2]
```diff
@@ -29,10 +29,19 @@ const endpoint: NextApiHandler = async (
...
             res.status(file.status);
-            res.setHeader('Content-Type', file.headers['content-type']);
+            const contentType = file.headers['content-type'];
+            res.setHeader(
+              'Content-Type',
+              typeof contentType === 'string'
+                ? contentType
+                : 'application/octet-stream',
+            );
+            const contentDisposition = file.headers['content-disposition'];
```
Comment (sam-drumm) [inline pages/api/reports/novalet/download/[fileName].tsx:39]: Just handling linting error 

```Argument of type 'AxiosHeaderValue | undefined' is not assignable to parameter of type 'string | number | readonly string[]'.
  Type 'undefined' is not assignable to type 'string | number | readonly string[]'.

File: __tests__/pages/api/auth/verify.spec.ts [hunk 3]
```diff
@@ -0,0 +1,209 @@
...
+const confirmVerifyCodeMock = confirmVerifyCode as jest.Mock;
+const setAuthCookieMock = setAuthCookie as jest.Mock;
+
+describe('POST /api/auth/verify', () => {
+  beforeEach(() => {
+    confirmVerifyCodeMock.mockReset();
+    setAuthCookieMock.mockReset();
+  });
+
+  it('returns 200 with the auth response when the body is a JSON string', async () => {
```
LBHTKarki, line 58: this is checking for correct JSON not just any JSON?

LBHTKarki (approved):

Duslerke (approved):

__tests__/pages/api/auth/verify.spec.ts
sam-drumm: Yeah checks it's been called with the shape of mockVerifyAuthRequest

PR #538, opened 2026-05-21
Bugfix/verify fail

sam-drumm:
What
--
On entering email or the verify code a 400 was being returned along with "must have email" response. Investigating this the body of the request was being sent as a buffer. I tried to set the header to json but still; 

Screenshot_2026-05-21_at_14 50 30-c772903c-1674-42

1. Browser sends application/json JSON body
2. API deliver req.body to app as a Buffer
3. Next req.body would deliver a Buffer-shaped object and the current handler can't handle it so email is undefined

So to solve this 

* Sets the header to application/json, doesn't change things but it is accurate.
* Update the handlers to parse buffer type

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -334,14 +334,14 @@ workflows:
               only: [development]
       - assume-role-staging:
           context: api-assume-role-housing-staging-context
-          requires:
-            - run-tests
-            - run-cypress-e2e
+          # requires:
+          #   - run-tests
+          #   - run-cypress-e2e
           filters:
             tags:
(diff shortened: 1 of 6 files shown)


Duslerke (approved):
What causes it to become a buffer?

sam-drumm:
> What causes it to become a buffer?

I think it's from serverless-http. But it has really confused me as this is not happening in production now and we are using serverless-http. This is being dumped in https://github.com/LBHackney-IT/lbh-housing-register/pull/532 so I haven't fully got to the bottom of that question.

PR #539, opened 2026-05-21
Development

sam-drumm:
What
--

Roll https://github.com/LBHackney-IT/lbh-housing-register/pull/538 into staging for testing

### DIFF
--- a/lib/store/auth.ts
+++ b/lib/store/auth.ts
@@ -16,6 +16,7 @@ export const createVerifyCode = createAsyncThunk(
 
     const res = await fetch(`/api/auth/generate`, {
       method: 'POST',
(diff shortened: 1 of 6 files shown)


lib/store/auth.ts
```diff
@@ -16,6 +16,7 @@ export const createVerifyCode = createAsyncThunk(
 
     const res = await fetch(`/api/auth/generate`, {
       method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
```
Duslerke, line 19: I recall you adding these, and then removing them the following day.
Should these be added back in?

Duslerke (approved):

PR #540, opened 2026-05-21
fix: remove headers

sam-drumm:
What
-- 

Remove headers introduced in https://github.com/LBHackney-IT/lbh-housing-register/pull/538 on the application fetches as they [break e2e tests](https://app.circleci.com/pipelines/github/LBHackney-IT/lbh-housing-register/1861/workflows/56fab1b8-8a3d-420d-aeda-b5139bee00e8/jobs/10103/tests). Auth fetches can keep the header as they handle all types.

I think this is what's happening when running locally; without the header, req.body is a string, so JSON.parse in routes like pages/api/applications/[id]/index.tsx works fine. With the header it parses req.body into an object, so JSON.parse throws a 500 ([object Object]).

The headers were not a requirement of that PR - they just seem sensible to keep as they reflect the request type. That header in theory doesn't do anything because the request is coming through as buffer on the deployed site anyway. But in the local env it's changing the request structure. 

Future
--

Best practice might be to set the header types across all requests and make sure the request is parsed in the e2e test properly (and in the routes). 

However serverless-http is being dropped in https://github.com/LBHackney-IT/lbh-housing-register/pull/532 which is the source of the buffer issue. So I think it's more valuable to address from that point onwards.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -291,7 +291,7 @@ workflows:
             tags:
               only: /^hackney-housing-register-v[0-9]+(\.[0-9]+)*$/
             branches:
-              only: [development, main, bugfix/failing-e2e]
+              only: [development, main]
           matrix:
             parameters:
               browser: [chrome, firefox]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -291,7 +291,7 @@ workflows:
             tags:
               only: /^hackney-housing-register-v[0-9]+(\.[0-9]+)*$/
             branches:
-              only: [development, main]
+              only: [development, main, bugfix/failing-e2e]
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -291,7 +291,7 @@ workflows:
             tags:
               only: /^hackney-housing-register-v[0-9]+(\.[0-9]+)*$/
             branches:
-              only: [development, main]
+              only: [development, main, bugfix/failing-e2e]
```
LBHTKarki: do you want to keep this?

sam-drumm: doh

LBHTKarki (approved):

PR #547, opened 2026-06-09
Ssi 1144 update GitHub actions

sam-drumm:
What
-- 

* Uses github action that supports node24 (See https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/)
* Uses updated orb (https://circleci.com/developer/orbs/orb/lbh-hackit/aws_assume_role) and removes need for python3 install
* Documents when the e2e tests run to match the pipeline strategy for integration testing


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hackney-housing-register",
-  "version": "2.2.0",
```
Duslerke, line 3: Any way to skip the release?

Pipeline and docs update is not something you'd want reflected in a release version. 🤔 
That's because nothing changes for the consumer of the package.

sam-drumm: The release package will also have all the new opennext build process not just this stuff. The version change is just main vs dev - main already at 2.2.1

Duslerke (approved):

PR #553, opened 2026-07-01
feat: protect applications route

sam-drumm:
## Summary

Restrict POST /api/applications to authenticated staff.
Validate auth email addresses on the generate and verify routes, mirroring the housing API.
Fix a failing sign-in Cypress test after the auth generate route started forwarding upstream status codes (#552).

## Context
Probely was hitting the public BFF POST /api/applications endpoint and creating DB records. Residents create applications via auth/generate, not this route — so blocking unauthenticated POSTs does not affect the resident journey. This PR adds the FE gate; the companion housing API [change](https://github.com/LBHackney-IT/lbh-housing-api/pull/268) adds the same check server-side.

### Cypress: sign-in error message
The E2E test "shows error message when create verify code API call fails" was failing:

Expected to find content: 'Unable to create verify code (500)' but never did.
The UI correctly showed Unable to create verify code (400).

The test was written before #552, when /api/auth/generate always returned 500 regardless of the housing API response. That route now forwards the upstream status code.

### Staff-only POST /api/applications
- Reject non-staff and read-only staff callers with 403 before parsing the request body.
- Forward the staff JWT to the housing API via authenticatedHousingAxios in addApplication.
- Updated unit tests in __tests__/pages/api/applications/index.spec.ts.

### Auth email validation 
- Added lib/utils/auth-email-validator.ts (mirrors housing API AuthEmailValidator rules).
- Applied validation on pages/api/auth/generate and pages/api/auth/verify — invalid emails return 400 with Email address is not valid.
- Unit tests for the validator and updated generate/verify route tests.

### Google Tag
- Tag value from SSM was not read at runtime, updates this so that it can be.

### CI
- Bumped CircleCI orbs (aws-cli, node, sonarcloud).

## Test plan
- [x]  Cypress sign-in spec — error message assertion passes with (400)
- [ ]  Unauthenticated POST /api/applications returns 403 
- [ ] Staff POST /api/applications still creates applications
- [ ] Auth generate/verify reject malformed or injection-style emails with 400


__tests__/pages/api/auth/generate.spec.ts
```diff
@@ -119,6 +119,21 @@ describe('POST', () => {
     },
   );
 
+  it('returns status code 400 without calling the backend when email format is invalid', async () => {
+    jsonParseSpy.mockReturnValueOnce({ email: "' OR 1=1 --" });
```
Duslerke: :D
Needs a semicolon though before the comment: `' OR 1=1; --`.

lib/utils/auth-email-validator.ts
Duslerke, line 1: Why not verify the email with regex pattern instead?
In my mind, there's no point in splitting the validation into individual email pieces if you're not going to use that to give more granular feedback to the user.

lib/utils/auth-email-validator.ts
```diff
@@ -0,0 +1,62 @@
...
+  const dotIndex = domainPart.indexOf('.');
+  if (dotIndex <= 0 || dotIndex >= domainPart.length - 1) {
+    return false;
+  }
+
+  return true;
+}
+
+export function isValidAuthEmail(email: unknown): email is string {
+  if (typeof email !== 'string' || email.trim() === '') {
```
Duslerke: Instead of allowing email to be `unknown` and then checking if it's a string, you can simply force it to be a string by the function typescript definition and ensure that all the callers of this func provide a string.

lib/utils/auth-email-validator.ts
```diff
@@ -0,0 +1,62 @@
...
+    return false;
+  }
+
+  const trimmed = email.trim();
```
Duslerke: `email.trim()` called twice. Once you get rid of `email` type check, you can move this to the top of func def and reference that within an if block checking if it's empty

lib/utils/auth-email-validator.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  if (containsDisallowedCharacter(trimmed)) {
+    return false;
+  }
+
+  if ((trimmed.match(/@/g) ?? []).length !== 1) {
+    return false;
+  }
+
+  return hasValidEmailFormat(trimmed);
```
Duslerke: This if block is arguably part of the `hasValidEmailFormat`. Not that there's any need to split this check out like that (see the file comment)

lib/utils/auth-email-validator.ts
sam-drumm: Yeah originally was 

const EMAIL_FORMAT = /^[^@\s]+@[^@\s]+\.[^@\s]+$/;

But Sonar flagged it for ReDoS risk

Duslerke: Doesn't this check get run on the client-side front-end?
If so, there's nothing for an attacked to DoS.

Duslerke: nvm, it's run on next.js server side `LBHackney-IT/lbh-housing-register/pages/api/auth/verify.tsx`

Duslerke (approved):

PR #554, opened 2026-07-03
fix: fix broken address validation

sam-drumm:
## Summary
Fixes the detached View Documents button on the person view page and completes the lbh-frontend 4.1 dialog migration. Removes the local Radix dialog wrapper and direct @radix-ui/react-dialog dependency in favour of the lbh-frontend/dialog subpath. Also adds address validation on the staff Add case form so missing address shows a proper GOV.UK error instead of failing silently and blocking adding a case.

### View Documents button
- Replaced invalid <a><button> markup with ButtonLink (target="_blank", rel="noreferrer") on the person view page so the evidence store link renders and behaves correctly.

### Dialog migration (lbh-frontend@4.1)
- Upgraded lbh-frontend to 4.1.x (Dialog with optional confirm/cancel actions + ./dialog export subpath).
- Removed local components/dialog.tsx and updated all dialog call sites to import Dialog from 'lbh-frontend/dialog'.
- Removed direct @radix-ui/react-dialog dependency (Radix now comes via lbh-frontend only).
- Added transpilePackages: ['lbh-frontend'] in next.config.js.
- Removed temporary dialog SCSS overrides and duplicate positioning rules — fixed upstream in lbh-frontend.
- Removed types/lbh-frontend.d.ts shim; package types are sufficient.

### Add case — address validation
- Address is validated on submit via Formik (stored outside form state).
- Shows “Address is a required field” in the error summary, with GOV.UK inline error styling on the Add address control.
- Error state clears when an address is saved (revalidation on addressHistory change).
- Added Cypress coverage for missing-address validation.

### Dependency cleanup
- Removed .npmrc legacy-peer-deps (no longer needed after lbh-frontend 4.x / React 19).

## Why
**View Documents**: Invalid HTML broke the link styling/behaviour in the browser.
**Dialog**: Avoid maintaining a duplicate Radix implementation; use the design system component and SSR-safe subpath import.
**Address validation**: Staff could fill the form, hit submit, and get no feedback when address was missing.

## Test plan
- [x] **View Documents** — open a person with evidence configured; button opens evidence store in a new tab.
- [x] **Dialogs** — smoke test: add/edit address modal, contact us, delete confirmation, sign-out warning, Novalet sync. Confirm close icon (top-right), overlay, and action footer layout.
- [x] **Add case** — submit without address → error summary + inline field error; add address → errors clear without re-submitting.
- [x] **Add case** — full happy path still saves successfully.
- [x] npm test / npm run build pass locally.
- [x] Cypress: add-case validation spec passes.

### Detached button
Screenshot 2026-07-02 at 12 33 37

### Detached dialog styling
Screenshot 2026-07-02 at 12 12 53

### Updated address validation
Screenshot 2026-07-02 at 12 27 03
Screenshot 2026-07-02 at 12 26 58


pages/applications/view/[id]/[person]/index.tsx
```diff
@@ -123,23 +123,22 @@ export default function ApplicationPersonPage({
...
-                    </a>
+                    {!userHasReadOnlyPermissionOnly && !evidenceLink ? (
+                      <ButtonLink
+                        href={`${evidenceLink}/deeplink?searchTerm=${fullName}&groupId=${cleanUpParams(
+                          applicant?.person?.id ?? '',
+                        )}&name=${fullName}&phone=${cleanUpParams(
+                          applicant?.contactInformation?.phoneNumber ?? '',
+                        )}&email=${cleanUpParams(
+                          applicant?.contactInformation?.emailAddress ?? '',
+                        )}`}
```
Comment (Duslerke) [inline pages/applications/view/[id]/[person]/index.tsx]: Extract the URL construction outside this href

components/button.tsx
```diff
@@ -87,6 +91,8 @@ export function ButtonLink({
       className={mergedClass}
       data-testid={dataTestId}
       draggable={false}
+      target={target}
```
Duslerke, line 94: What is a `target`?
What do we need `rel` and `target` for? Why isn't `href` enough?

components/admin/MainApplicantForm.tsx
```diff
@@ -81,6 +91,20 @@ const ethnicityOtherSection = getSectionData(
...
 
+function RevalidateOnAddressHistoryChange({
+  addressHistory,
+}: {
+  addressHistory: Address[];
+}) {
+  const { validateForm } = useFormikContext();
+
+  useEffect(() => {
+    void validateForm();
```
Duslerke: See the sonar warning

components/button.tsx
sam-drumm: In the case of view documents that was broken;
  href = url
  target="_blank" (new tab/window)
  rel="noreferrer" (saftey limits the new window from using stuff like window.opener on the source window)
  
 href alone would just open in the same window.

Duslerke (approved):

PR #562, opened 2026-07-28
chore: security deps and gtm build

sam-drumm:
## What

Clears a high audit finding, fixes Google Tag Manager not rendering on production after the OpenNext migration, and cleans up the environment variables around Sentry and evidence-store config that got confusing during the GTM investigation.

## Security deps

- jest-junit 16 → 17 (clears moderate uuid advisory)
- sass 1.98 → 1.102 (pulls immutable@5.1.9, clears high immutable advisories)

## GTM
GTM_ID was read from the Lambda environment in getInitialProps, but pages without getServerSideProps are statically optimized at next build time in CI — long before serverless deploy sets that Lambda env var. The static HTML never saw a value.

Fixed by reading NEXT_PUBLIC_GTM_ID directly (through housing-register-fe-build-context) and removing the now-unused GTM_ID from serverless.yml.

## Env var cleanup
Auditing which vars actually need NEXT_PUBLIC_ surfaced a few that were mislabelled or misplaced:

- NEXT_PUBLIC_EVIDENCE_STORE → EVIDENCE_STORE_URL. Every usage reads it inside getServerSideProps and passes it down as a prop — nothing reads process.env client-side. The NEXT_PUBLIC_ prefix was pointless.
- Sentry environment split into two vars that were previously sharing NEXT_PUBLIC_ENV:
  - sentry.server.config.js / sentry.edge.config.js now read SENTRY_ENVIRONMENT, set at runtime via serverless.yml (${opt:stage}) — correct, since server/edge code runs in the Lambda and can read real runtime env.
  - instrumentation-client.ts (browser) keeps NEXT_PUBLIC_ENV, since only build-time inlining can get a value into client code. This is now explicitly derived in the CircleCI build job (see below) instead of coming from serverless.yml, which never actually reached it.
- Added a !isLocalHost guard to the client Sentry init, so it can't activate against localhost/127.0.0.1/::1 — protects local dev and the Cypress e2e job (which serves the deploy-targeted build from a local next start and would otherwise start reporting/console-capturing on every test run).
- Removed sentry.properties. It's dead config
- next.config.js: silent: true → silent: !process.env.CI (default). Source map upload was failing silently either way — this makes upload/warning logs visible in CI without adding noise to local dev builds.
- .env.sample documents both Sentry vars for local dev.

## CircleCI

- Attached housing-register-fe-build-context to the build job itself. It was previously only on the deploy-to-* jobs — meaning NEXT_PUBLIC_GTM_ID and the Sentry source-map vars (SENTRY_AUTH_TOKEN/SENTRY_ORG/SENTRY_PROJECT) never reached the actual next build step where they're needed.
- build job now resolves the deploying stage from CIRCLE_BRANCH/CIRCLE_TAG before next build, so NEXT_PUBLIC_ENV (and therefore client Sentry's environment tag and sample rate) is correct per stage instead of always undefined. Mapping mirrors the deploy-to-* job filters exactly (tag → production, main → staging, development → development).

## Test plan

- [ ] Confirm CI build log shows export NEXT_PUBLIC_ENV=<stage> per branch/tag
- [ ] Confirm CI build log no longer shows "No auth token provided" (or shows a source map upload summary instead)
- [ ] Deployed development/staging/production: GTM fires, client + server Sentry events tagged with correct environment
- [x] Cypress e2e run: no Sentry events/console noise from the e2e job


serverless.yml
```diff
@@ -59,7 +59,6 @@ functions:
       NOTIFY_TEMPLATE_DISQUALIFY: ${ssm:arn:aws:ssm:eu-west-2:${param:centralAccount}:parameter/housing-${self:provider.stage}/housing-register-${self:provider.stage}/housing-register/notify-template-disqualify}
       NOTIFY_TEMPLATE_MEDICAL_NEED: ${ssm:arn:aws:ssm:eu-west-2:${param:centralAccount}:parameter/housing-${self:provider.stage}/housing-register-${self:provider.stage}/housing-register/notify-template-medical-need}
       NEXT_PUBLIC_EVIDENCE_STORE: ${ssm:arn:aws:ssm:eu-west-2:${param:centralAccount}:parameter/housing-${self:provider.stage}/housing-register-${self:provider.stage}/housing-register/evidence-store-url}
-      GTM_ID: ${ssm:arn:aws:ssm:eu-west-2:${param:centralAccount}:parameter/housing-${self:provider.stage}/housing-register-${self:provider.stage}/housing-register/gtm-id}
```
Duslerke, line 62: Where does the `NEXT_PUBLIC_GTM_ID` get ingested from?

pages/_document.tsx
```diff
@@ -1,31 +1,12 @@
...
-  static async getInitialProps(
-    ctx: DocumentContext,
-  ): Promise<
-    AppDocumentProps & Awaited<ReturnType<typeof Document.getInitialProps>>
-  > {
-    const initialProps = await Document.getInitialProps(ctx);
-
-    return { ...initialProps, gtmId: process.env.GTM_ID };
-  }
+const gtmId = process.env.NEXT_PUBLIC_GTM_ID;
```
Duslerke, line 7: It's not listed in `serverless.yml`, nor in circleci config

serverless.yml
sam-drumm: circle at buildtime

sam-drumm: i.e. so it's on every page regardless of static or not

Duslerke (dismissed):
The `GTM_ID` comes from CircleCI's this particular project settings.
It's not on AWS to be needed to get pulled.

Duslerke (dismissed):

Miles-Alford (approved):

PR #564, opened 2026-07-29
fix: require authentication on /api/notify/[template] and stop trusting client-supplied email data

sam-drumm:
## What

- The endpoint now requires a valid resident session (getUser(req)?.application_id) and returns 401 without contacting Notify if there isn't one.
- It looks up the caller's own application server-side (404 if not found) and builds the entire NotifyRequest from that trusted record via new lib/utils/notifyRequestBuilders.ts helpers.
- The request body is no longer parsed or used at all — a caller can no longer influence the email address, reference, or personalisation sent to Notify.
- Simplified the frontend (sendConfirmation/sendMedicalNeed/sendDisqualifyEmail thunks and their call sites in declaration.tsx/summary.tsx) to POST with no body, since the server derives everything itself now.

## Testing

- Unit tests (__tests__/pages/api/notify/[template].spec.ts, lib/utils/notifyRequestBuilders.test.ts): auth rejection, 404 handling, per-template request building, and that the request body is never parsed/used.
- E2E: fixed two existing Cypress specs whose backend mocks were sized for the old call count and would have silently stopped exercising the Notify flow once this endpoint started looking up the application itself. Added a new spec proving, against a real running server, that (a) an unauthenticated request never reaches Notify, and (b) a spoofed request body is ignored in favour of the resident's real application data.

### DIFF
--- a/__tests__/pages/api/notify/[template].spec.ts
+++ b/__tests__/pages/api/notify/[template].spec.ts
@@ -1,0 +1,253 @@
+/**
+ * @jest-environment node
+ */
+
+import { StatusCodes } from 'http-status-codes';
+import { createMocks, RequestMethod, RequestOptions } from 'node-mocks-http';
+
+import { Application } from '../../../../domain/HousingApi';
+import { NotifyRequest, NotifyResponse } from '../../../../domain/govukNotify';
+import endpoint from '../../../../pages/api/notify/[template]';
+import { ApiRequest, ApiResponse } from '../../../../testUtils/types';
+
+jest.mock('../../../../lib/utils/users', () => ({
+  getUser: jest.fn(),
+}));
+
+jest.mock('../../../../lib/gateways/applications-api', () => ({
+  getApplication: jest.fn(),
+}));
+
+jest.mock('../../../../lib/gateways/notify-api', () => ({
+  sendNewApplicationEmail: jest.fn(),
+  sendMedicalNeedEmail: jest.fn(),
+  sendDisqualifyEmail: jest.fn(),
+}));
+
+jest.mock('../../../../lib/utils/notifyRequestBuilders', () => ({
+  buildNewApplicationNotifyRequest: jest.fn(),
+  buildMedicalNeedNotifyRequest: jest.fn(),
+  buildDisqualifyNotifyRequest: jest.fn(),
+}));
+
+import { getApplication } from '../../../../lib/gateways/applications-api';
+import {
+  sendDisqualifyEmail,
+  sendMedicalNeedEmail,
+  sendNewApplicationEmail,
+} from '../../../../lib/gateways/notify-api';
+import {
+  buildDisqualifyNotifyRequest,
+  buildMedicalNeedNotifyRequest,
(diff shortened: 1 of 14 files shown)


__tests__/pages/api/notify/[template].spec.ts
```diff
@@ -0,0 +1,253 @@
...
+    sendDisqualifyEmailMock.mockResolvedValue(mockNotifyResponse);
+  });
+
+  it('returns 400 for non-POST methods without checking the session', async () => {
+    const { req, res } = mockRequest('new-application', 'GET');
+
+    await endpoint(req, res);
+
+    expect(getUserMock).not.toHaveBeenCalled();
+    expect(res.statusCode).toBe(StatusCodes.BAD_REQUEST);
```
Comment (LBHTKarki) [inline __tests__/pages/api/notify/[template].spec.ts]: Would 405 (method not allowed) be more appropriate here? Would tie in with the message that tells it failed beacuse of invalid method.

__tests__/pages/api/notify/[template].spec.ts
```diff
@@ -0,0 +1,253 @@
...
+  it('returns 500 when the Notify send fails', async () => {
+    sendDisqualifyEmailMock.mockRejectedValue(new Error('notify down'));
+
+    const { req, res } = mockRequest('disqualify');
+
+    await endpoint(req, res);
+
+    expect(res.statusCode).toBe(StatusCodes.INTERNAL_SERVER_ERROR);
+    expect(res._getJSONData()).toStrictEqual({
+      message: 'Unable to send email',
```
Comment (LBHTKarki) [inline __tests__/pages/api/notify/[template].spec.ts:272]: for debugging purposes would it be useful to specify where the error is i.e. notify vs app?

lib/utils/notifyRequestBuilders.ts
```diff
@@ -0,0 +1,60 @@
...
+// These builders derive the entire NotifyRequest from the application record
+// fetched server-side, rather than trusting client-supplied email/reference/
+// personalisation values - the caller only chooses which template to send,
+// authorised against their own application.
+function baseNotifyRequest(
+  application: Application,
+): Pick<NotifyRequest, 'emailAddress' | 'reference'> {
+  return {
+    emailAddress:
+      application.mainApplicant?.contactInformation?.emailAddress ?? '',
```
LBHTKarki, line 16: I guess it's not possible for the email to be null?

__tests__/pages/api/notify/[template].spec.ts
Comment (sam-drumm) [reply __tests__/pages/api/notify/[template].spec.ts]: Will also backlog because all the endpoints just have a generic 400

lib/utils/notifyRequestBuilders.ts
sam-drumm: Actually it is with but would be captured on the '' fallback and notify should reject it. I've updated to reject null, empty, or whitespace-only email addresses with "Application has no email address" before notify.

LBHTKarki (dismissed):

LBHTKarki (dismissed):

LBHTKarki (dismissed):

LBHTKarki (approved):

PR #565, opened 2026-07-30
fix: guard checkbox-list data against unanswered questions

sam-drumm:
## What
[Sentry errors](https://london-borough-of-hackney.sentry.io/issues/7641967071/?alert_rule_id=15956553&alert_type=issue&environment=production&notification_uuid=8a33aca4-29ca-4cfe-8bcd-30d744377984&project=6292602&referrer=issue_alert-slack) unearthed that staff can't open /applications/view/[id]/[person] for certain applicants — the page throws an unhandled 500 on every load:

TypeError: Cannot read properties of null (reading 'toString')
    at .../pages/applications/view/[id]/[person].js

We're aware the form itself has historical issues outside the scope of maintenance - so it might also be released to the expectation that values are never null but the reality being different. In some places fields are sometimes not guarded correctly (both in terms of user messaging and rejection criteria). 

## Cause
Form answers are persisted as JSON.stringify(value) (lib/store/applicant.ts). An unanswered field is stored as the string "null". Several places in lib/utils/applicationQuestions.ts and lib/utils/checkboxListData.ts checked truthiness of that stored string before parsing it, then called a method on the parsed result without checking for null:

- getQuestionValue: capitalize(jsonParse(questionValue).toString()) → null.toString() throws.
- residentialStatusCheckboxList: institutionsArray.map(...) on a parsed null → same crash.
- addressHistoryCheckboxList: addressHistorysArray.map(...) on a parsed null → same crash.
- ethnicityExtendedCategoryString: silently returned undefined (rendering literal "undefined" text), and separately .filter(...)[0].label could throw on an unrecognised/legacy stored value.

Since getQuestionValue is called multiple times across this file for practically every field on the page, any applicant with even one unanswered question can crash the entire page for every member of staff who opens it.

residentialStatusCheckboxList was calling questionLookup(...) for institutions without passing applicant — a separate, pre-existing bug that meant institutions always silently rendered 'N/A' regardless of real data. 

## Fix

Check the parsed value for null/undefined/empty rather than trusting the raw string's truthiness, and use Array.isArray(...) / .find()?.label ?? 'N/A' instead of unconditional .map/.filter()[0] in the affected spots.

## Tests

- lib/utils/applicationQuestions.test.ts (new, 10 tests) and lib/utils/checkboxListData.test.ts (new, 11 tests) — cover the "null"-literal case, empty/undefined, and the happy path for each fixed function.
- cypress/e2e/pages/reviewApplicant.cy.ts — 4 new e2e tests reproducing each crash site end-to-end against a mocked application with a "null"-literal answer, asserting the page loads and shows N/A instead of a 500.

### DIFF
--- a/lib/utils/applicationQuestions.ts
+++ b/lib/utils/applicationQuestions.ts
@@ -16,7 +16,7 @@ export const questionLookup = (
   return applicant?.questions?.find((q) => q.id === questionId)?.answer;
 };
 
-export const jsonParse = (parseItem: string): string => {
+export const jsonParse = (parseItem: string): unknown => {
   try {
     return JSON.parse(parseItem);
   } catch {
@@ -29,7 +29,15 @@ export const getQuestionValue = (
   applicant?: Applicant,
 ): string => {
   const questionValue = questionLookup(questionId, applicant);
-  return questionValue
-    ? capitalize(jsonParse(questionValue).toString())
-    : 'N/A';
+  if (questionValue === undefined) {
+    return 'N/A';
+  }
+
+  // Answers are stored via JSON.stringify, so an unanswered question can be
+  // persisted as the *string* "null" (JSON.stringify(null)) - that string is
+  // truthy, but parses back to the JS value `null`, which has no .toString().
+  const parsedValue = jsonParse(questionValue);
+  return parsedValue === null || parsedValue === undefined || parsedValue === ''
+    ? 'N/A'
+    : capitalize(String(parsedValue));
 };
--- a/lib/utils/applicationQuestions.ts
+++ b/lib/utils/applicationQuestions.ts
@@ -16,7 +16,14 @@ export const questionLookup = (
   return applicant?.questions?.find((q) => q.id === questionId)?.answer;
 };
 
-export const jsonParse = (parseItem: string): unknown => {
+export const jsonParse = (parseItem: string | undefined | null): unknown => {
+  // Question.answer is optional on the domain type, and callers previously
+  // papered over that with `answer!` before handing it to JSON.parse - which
+  // would coerce undefined to the string "undefined" and throw a SyntaxError.
+  if (parseItem == null || parseItem === '') {
+    return '';
+  }
+
   try {
     return JSON.parse(parseItem);
   } catch {
--- a/lib/utils/checkboxListData.ts
+++ b/lib/utils/checkboxListData.ts
@@ -3,6 +3,7 @@ import { CheckBoxListPageProps } from '../../components/admin/checkbox-list';
 import {
   questionLookup,
   getQuestionValue,
+  jsonParse,
 } from '../../lib/utils/applicationQuestions';
 import { QuestionKey } from './question-data';
 import {
@@ -121,7 +122,7 @@ const ethnicityExtendedCategoryString = (applicant?: Applicant) => {
   const JsonStringEthnicityCategory =
     questionLookup(QuestionKey.ETHNICITY_MAIN_CATEGORY, applicant) || '""';
 
-  const ethnicityCategory = JSON.parse(JsonStringEthnicityCategory);
+  const ethnicityCategory = jsonParse(JsonStringEthnicityCategory);
 
   if (!ethnicityCategory) {
     return 'N/A';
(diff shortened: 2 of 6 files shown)


lib/utils/checkboxListData.ts
```diff
@@ -140,15 +140,22 @@ const ethnicityExtendedCategoryString = (applicant?: Applicant) => {
...
+  // returns the JS value `null`.
   const ethnicityExtendedCategory = JSON.parse(
     ethnicityExtendedCategoryQuestion.answer!,
   );
 
-  if (ethnicityExtendedCategory) {
-    return ethnicityCategoryOptions.filter(
-      (option) => option.value === ethnicityExtendedCategory,
-    )[0].label;
+  if (!ethnicityExtendedCategory) {
```
Duslerke, line 152: In response to code comment ln143-145 
The `ethnicityExtendedCategoryQuestion.answer!` has a `!` though... doesn't it mean it's guaranteed to not be null?

lib/utils/checkboxListData.ts
```diff
@@ -307,12 +314,16 @@ export const residentialStatusCheckboxList = (
...
 
   let institutionsText = '';
-  if (institutions) {
-    const institutionsArray = JSON.parse(institutions);
-    institutionsArray.map((item: string) => {
+  // Answers are stored via JSON.stringify, so an unanswered question is
+  // persisted as the string "null" - truthy as a string, but JSON.parse
+  // returns the JS value `null`, not an array.
+  const institutionsArray = institutions ? JSON.parse(institutions) : null;
+  if (Array.isArray(institutionsArray) && institutionsArray.length > 0) {
```
Duslerke, line 327: I wonder if you even need the `institutionsArray.length > 0` condition
as for each should simply do 0 iterations when array has 0 items. Well in C# at least... don't recall how it's in JS

Duslerke (dismissed):

lib/utils/applicationQuestions.ts
```diff
@@ -16,7 +16,14 @@ export const questionLookup = (
...
 };
 
-export const jsonParse = (parseItem: string): string => {
+export const jsonParse = (parseItem: string | undefined | null): unknown => {
+  // Question.answer is optional on the domain type, and callers previously
+  // papered over that with `answer!` before handing it to JSON.parse - which
+  // would coerce undefined to the string "undefined" and throw a SyntaxError.
+  if (parseItem == null || parseItem === '') {
+    return '';
+  }
```
Duslerke, line 25: This does not handle the `undefined`. So how come not?

```suggestion
export const jsonParse = (parseItem: string | undefined | null): unknown => {
  // Question.answer is optional on the domain type, and callers previously
  // papered over that with `answer!` before handing it to JSON.parse - which
  // would coerce undefined to the string "undefined" and throw a SyntaxError.
  if (!parseItem) {
    return null;
  }
```

This would allow you to change further code utilising this func from:
`const institutionsArray = institutions ? jsonParse(institutions) : null;`

into:
`const institutionsArray = jsonParse(institutions);`

lib/utils/applicationQuestions.ts
```diff
@@ -29,7 +36,15 @@ export const getQuestionValue = (
...
+    return 'N/A';
+  }
+
+  // Answers are stored via JSON.stringify, so an unanswered question can be
+  // persisted as the *string* "null" (JSON.stringify(null)) - that string is
+  // truthy, but parses back to the JS value `null`, which has no .toString().
+  const parsedValue = jsonParse(questionValue);
+  return parsedValue === null || parsedValue === undefined || parsedValue === ''
+    ? 'N/A'
+    : capitalize(String(parsedValue));
```
Duslerke, line 49: Then if the above comment is addressed, this becomes:

```suggestion
  return capitalize(String(jsonParse(questionValue) || 'N/A'));
```

lib/utils/applicationQuestions.ts
sam-drumm: It does handle undefined but yours is clearer. 

== will handle null and undefined as the same aka 'nullish'

lib/utils/applicationQuestions.ts
sam-drumm: In your suggestion I think 'false' would resolve to N/A which we don't want. 

i.e. if the stored answer is "false" > JSON.parse will resolve to false which will fall to 'N/A'. Or "0" would do the same but not in the previous.

Duslerke: Woops, I foolishly forgot about that. Let me correct this:

```suggestion
  return capitalize(String(jsonParse(questionValue) ?? 'N/A'));
```

Duslerke: You'd need to bump this though to at least `ES2020`:
https://github.com/LBHackney-IT/lbh-housing-register/blob/main/tsconfig.json#L3

sam-drumm: I don't think we need to bump for that feature ( I will update that and browserlist though separately), but I'm just thinking how will '' be treated? 

capitalise(String('')) -> ''

Previous '' would be N/A

sam-drumm: I guess jsonParse should return null for missing values just not sure what the flow on of that change will be - might need to deal with that separately

Duslerke (approved):

lib/utils/applicationQuestions.ts
Duslerke: > String(''))

It would not become empty string after json parse as the empty string is falsy.

Duslerke: > I guess jsonParse should return null for missing values just not sure what the flow on of that change will be - might need to deal with that separately

It's fine. If it's not a quick change. It can be left for when you have time for this. It's aesthetics anyway.

PR #570, opened 2026-08-03
Fix/api method not allowed

sam-drumm:
## Summary
After adding 405 to the Notifiy endpoint as part of #564 - this was a clean-up task for all API endpoints returning 400 Bad Request for unsupported HTTP methods (should be 405 Method Not Allowed). On auditing every endpoint's error handling several related bugs where errors weren't being handled consistently or correctly.

## Method handling
All API routes now return 405 Method Not Allowed with an Allow header for unsupported methods, instead of 400 Bad Request:

- applications (GET, POST), applications/[id] (PATCH), applications/[id]/evidence, applications/[id]/note, applications/[id]/complete, address/[postcode], auth/generate, auth/verify, auth/exit, reports/internal/download, reports/novalet/generate, reports/novalet/approve/[fileName], reports/novalet/download/[fileName]
- admin/logout had no default case at all, so non-GET requests silently returned an empty 200 OK. It now returns 405 like the rest.
## Error handling bugs

- Requests that could hang with no response. applications/[id] swallowed the "no response received" / "request setup failed" with only a console.log, never writing a status. reports/novalet/generate and reports/novalet/approve/[fileName] didn't wrap their gateway calls in try/catch at all, so a thrown error left the request hanging. All now guarantee a response on every path.
- Malformed JSON bodies returned 500 instead of 400. applications, applications/[id], applications/[id]/evidence, and applications/[id]/note all did JSON.parse(req.body) with no error handling, so a bad request body was reported to Sentry as a server error. Parsing is now wrapped separately and returns 400 Bad Request / "Unable to parse request".
- Upstream error responses were being collapsed to a generic 500. Several endpoints didn't forward the backend API's actual status code/body on failure. Now consistently checking axios.isAxiosError(error) && error.response and forwarding it, falling back to 500 only for genuine non-HTTP failures.
- Internal details leaking in error responses. reports/internal/download was returning the raw Axios response object and error in the JSON body on failure. Now just logs server-side and returns a generic message.
- Notify gateway silently swallowing failures. lib/gateways/notify-api.ts's sendEmail caught Notify API errors, reported them to Sentry, but then resolved successfully with undefined - so /api/notify/[template] always returned 200 OK regardless of whether the email actually sent. It also console.logged the successful response instead of returning response.data, so a real success also came back as undefined. Both now behave correctly: success returns the actual NotifyResponse, failure rethrows so the endpoint can respond with the right status.
- Sentry noise from scanner traffic. Next's own body parser throws Invalid JSON/Invalid body/Body exceeded ... limit before our route code runs (already handled correctly by Next with 400/413). These were filling Sentry with noise from Probely scans; now filtered via ignoreErrors.
## Other fixes found along the way

- Cleaned up confirm dialogs on the staff "remove household member" action. Surfaced from a cypress test flake in the local suite. components/admin/other-members.tsx rendered a delete-confirmation dialog per row, all driven by one shared piece of state - deleting a member with siblings in the household opened multiple modal dialogs on top of each other, and only the topmost was interactive. Moved to a single dialog per table.
- sendDisqualifyEmail/sendConfirmation/sendMedicalNeed thunk failures were completely invisible. These fetch /api/notify/[template] and deliberately don't block the resident's flow on failure, but nothing logged a rejection either - a 401/404/422/500 from Notify vanished with no trace. Added console.error before rejectWithValue.

## Test plan

- [x] tsc --noEmit, ESLint, and full Jest suite (28 suites / 260 tests) all pass
- [x] Added/updated unit tests for every endpoint's new status codes and error paths (__tests__/pages/api/**, __tests__/lib/gateways/notify-api.spec.ts)
- [x] Reviewed and updated the E2E suite for the new response codes (cypress/e2e/pages/api/notify/[template].cy.ts, declaration.cy.ts, managerActions.cy.ts)
- [x] Run full Cypress E2E suite before merge

### DIFF
--- a/lib/gateways/notify-api.ts
+++ b/lib/gateways/notify-api.ts
@@ -6,55 +6,51 @@ import { NotifyClient } from 'notifications-node-client';
 // itself (rather than relying on `console.error` + breadcrumbs), so the
 // payload that caused a Notify failure is always visible on the event,
 // regardless of breadcrumb/scope isolation health.
-function captureNotifyError(err: unknown, request: NotifyRequest) {
+function captureNotifyError(err: unknown, request: NotifyRequest): void {
   Sentry.captureException(err, { extra: { notifyRequest: request } });
 }
 
-export const sendNewApplicationEmail = async (
+async function sendEmail(
+  templateId: string | undefined,
   request: NotifyRequest,
-): Promise<NotifyResponse> => {
+): Promise<NotifyResponse> {
   const notifyClient = new NotifyClient(process.env.NOTIFY_API_KEY);
-  const response = await notifyClient
-    .sendEmail(
-      process.env.NOTIFY_TEMPLATE_NEW_APPLICATION,
+
+  try {
+    const response = await notifyClient.sendEmail(
+      templateId,
       request.emailAddress,
       {
         personalisation: request.personalisation,
         reference: request.reference,
       },
-    )
-    .then((response: NotifyResponse) => console.log(response))
-    .catch((err: NotifyResponse) => captureNotifyError(err, request));
+    );
 
-  return response as NotifyResponse;
-};
+    // sendEmail resolves with the raw Axios response - the NotifyResponse
+    // shape (id/reference/content) lives on `.data`.
+    return response.data as NotifyResponse;
+  } catch (err) {
+    // Previously this was swallowed here (logged/captured then treated as a
+    // successful, empty response), so the caller always got a 200 with no
+    // body regardless of whether the email actually sent. Capture to Sentry
+    // with the payload attached, then rethrow so /api/notify/[template] can
+    // return a proper error status to the client.
+    captureNotifyError(err, request);
+    throw err;
+  }
+}
 
-export const sendMedicalNeedEmail = async (
+export const sendNewApplicationEmail = (
   request: NotifyRequest,
-): Promise<NotifyResponse> => {
-  const notifyClient = new NotifyClient(process.env.NOTIFY_API_KEY);
-  const response = await notifyClient
-    .sendEmail(process.env.NOTIFY_TEMPLATE_MEDICAL_NEED, request.emailAddress, {
-      personalisation: request.personalisation,
-      reference: request.reference,
-    })
-    .then((response: NotifyResponse) => console.log(response))
-    .catch((err: NotifyResponse) => captureNotifyError(err, request));
+): Promise<NotifyResponse> =>
+  sendEmail(process.env.NOTIFY_TEMPLATE_NEW_APPLICATION, request);
 
-  return response as NotifyResponse;
-};
-
-export const sendDisqualifyEmail = async (
+export const sendMedicalNeedEmail = (
   request: NotifyRequest,
-): Promise<NotifyResponse> => {
-  const notifyClient = new NotifyClient(process.env.NOTIFY_API_KEY);
-  const response = await notifyClient
-    .sendEmail(process.env.NOTIFY_TEMPLATE_DISQUALIFY, request.emailAddress, {
-      personalisation: request.personalisation,
-      reference: request.reference,
-    })
-    .then((response: NotifyResponse) => console.log(response))
-    .catch((err: NotifyResponse) => captureNotifyError(err, request));
+): Promise<NotifyResponse> =>
+  sendEmail(process.env.NOTIFY_TEMPLATE_MEDICAL_NEED, request);
 
-  return response as NotifyResponse;
-};
+export const sendDisqualifyEmail = (
+  request: NotifyRequest,
+): Promise<NotifyResponse> =>
+  sendEmail(process.env.NOTIFY_TEMPLATE_DISQUALIFY, request);
--- a/pages/api/address/[postcode]/index.ts
+++ b/pages/api/address/[postcode]/index.ts
@@ -7,29 +7,27 @@ const endpoint: NextApiHandler = async (
   req: NextApiRequest,
   res: NextApiResponse,
 ) => {
-  switch (req.method) {
-    case 'GET':
-      try {
-        const { postcode } = req.query;
-        if (!postcode) {
-          res.status(StatusCodes.BAD_REQUEST).send('Missing postcode');
-          return;
-        }
-        const data = await lookUpAddress(postcode);
-        res.status(StatusCodes.OK).json(data);
-      } catch (error) {
-        console.error(error);
-        res
-          .status(StatusCodes.INTERNAL_SERVER_ERROR)
-          .json({ message: 'Unable to look up address' });
-      }
-      break;
+  if (req.method !== 'GET') {
+    res
+      .setHeader('Allow', 'GET')
+      .status(StatusCodes.METHOD_NOT_ALLOWED)
+      .json({ message: 'Method not allowed' });
+    return;
+  }
 
-    default:
-      res
-        .setHeader('Allow', 'GET')
-        .status(StatusCodes.METHOD_NOT_ALLOWED)
-        .json({ message: 'Method not allowed' });
+  try {
+    const { postcode } = req.query;
+    if (!postcode) {
+      res.status(StatusCodes.BAD_REQUEST).send('Missing postcode');
+      return;
+    }
+    const data = await lookUpAddress(postcode);
+    res.status(StatusCodes.OK).json(data);
+  } catch (error) {
+    console.error(error);
+    res
+      .status(StatusCodes.INTERNAL_SERVER_ERROR)
+      .json({ message: 'Unable to look up address' });
   }
 };
 
--- a/pages/api/address/[postcode]/index.ts
+++ b/pages/api/address/[postcode]/index.ts
@@ -27,8 +27,9 @@ const endpoint: NextApiHandler = async (
 
     default:
       res
-        .status(StatusCodes.BAD_REQUEST)
-        .json({ message: 'Invalid request method' });
+        .setHeader('Allow', 'GET')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
   }
 };
 
--- a/pages/api/admin/logout.tsx
+++ b/pages/api/admin/logout.tsx
@@ -7,29 +7,27 @@ const endpoint: NextApiHandler = async (
   req: NextApiRequest,
   res: NextApiResponse,
 ) => {
+  if (req.method !== 'GET') {
+    res
+      .setHeader('Allow', 'GET')
+      .status(StatusCodes.METHOD_NOT_ALLOWED)
+      .json({ message: 'Method not allowed' });
+    return;
+  }
+
   const isAdminUser = !!getSession(req);
 
-  switch (req.method) {
-    case 'GET':
-      try {
-        if (isAdminUser) {
-          removeHackneyToken(res);
-        }
+  try {
+    if (isAdminUser) {
+      removeHackneyToken(res);
+    }
 
-        res.status(StatusCodes.OK).json({ message: 'Admin sign out' });
-      } catch (error) {
-        console.error(error);
-        res
-          .status(StatusCodes.INTERNAL_SERVER_ERROR)
-          .json({ message: 'Unable to sign out' });
-      }
-      break;
-
-    default:
-      res
-        .setHeader('Allow', 'GET')
-        .status(StatusCodes.METHOD_NOT_ALLOWED)
-        .json({ message: 'Method not allowed' });
+    res.status(StatusCodes.OK).json({ message: 'Admin sign out' });
+  } catch (error) {
+    console.error(error);
+    res
+      .status(StatusCodes.INTERNAL_SERVER_ERROR)
+      .json({ message: 'Unable to sign out' });
   }
 };
 
--- a/pages/api/admin/logout.tsx
+++ b/pages/api/admin/logout.tsx
@@ -24,6 +24,12 @@ const endpoint: NextApiHandler = async (
           .json({ message: 'Unable to sign out' });
       }
       break;
+
+    default:
+      res
+        .setHeader('Allow', 'GET')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
   }
 };
 
--- a/pages/api/applications/[id]/complete.tsx
+++ b/pages/api/applications/[id]/complete.tsx
@@ -9,36 +9,32 @@ const endpoint: NextApiHandler = async (
   req: NextApiRequest,
   res: NextApiResponse,
 ) => {
-  switch (req.method) {
-    case 'PATCH':
-      try {
-        const id = req.query.id as string;
+  if (req.method !== 'PATCH') {
+    res
+      .setHeader('Allow', 'PATCH')
+      .status(StatusCodes.METHOD_NOT_ALLOWED)
+      .json({ message: 'Method not allowed' });
+    return;
+  }
 
-        if (canUpdateApplication(req, id)) {
-          const data = await completeApplication(id, req);
-          res.status(StatusCodes.OK).json(data);
-        } else {
-          res
-            .status(StatusCodes.FORBIDDEN)
-            .json({ message: 'Unable to update application' });
-        }
-      } catch (error) {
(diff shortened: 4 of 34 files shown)


pages/api/reports/novalet/generate.tsx
```diff
@@ -23,27 +23,35 @@ const endpoint: NextApiHandler = async (
...
 
       break;
 
     default:
       res
-        .status(StatusCodes.BAD_REQUEST)
-        .json({ message: 'Invalid request method' });
+        .setHeader('Allow', 'POST')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
```
Duslerke: A cleaner way to do this would have been to make this a guard condition and drop the switch case.

pages/api/reports/novalet/download/[fileName].tsx
```diff
@@ -60,8 +60,9 @@ const endpoint: NextApiHandler = async (
 
     default:
       res
-        .status(StatusCodes.BAD_REQUEST)
-        .json({ message: 'Invalid request method' });
+        .setHeader('Allow', 'GET')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
```
Comment (Duslerke) [inline pages/api/reports/novalet/download/[fileName].tsx]: Switch only has one entry, so this `default` block can simply be a guard clause and you get to drop the extra indent.

pages/api/reports/novalet/approve/[fileName].tsx
```diff
@@ -23,30 +23,44 @@ const endpoint: NextApiHandler = async (
...
+            }
           } else {
-            res.send({
-              message: 'Unable to approve export file',
-            });
+            // Previously fell through here with no response ever written,
+            // leaving the caller's request to hang.
+            res
+              .status(StatusCodes.INTERNAL_SERVER_ERROR)
+              .json({ message: 'Unable to approve export file' });
```
Comment (Duslerke) [inline pages/api/reports/novalet/approve/[fileName].tsx]: Should this get logged?
I image if the sub-request fell through, we would want to know if it's something we have any control over.

pages/api/reports/novalet/approve/[fileName].tsx
```diff
@@ -23,30 +23,44 @@ const endpoint: NextApiHandler = async (
...
 
       break;
 
     default:
       res
-        .status(StatusCodes.BAD_REQUEST)
-        .json({ message: 'Invalid request method' });
+        .setHeader('Allow', 'POST')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
```
Comment (Duslerke) [inline pages/api/reports/novalet/approve/[fileName].tsx]: As before, the switch case has only 1 entry, so it can be dropped ... reducing the indent.

pages/api/reports/novalet/generate.tsx
```diff
@@ -23,27 +23,35 @@ const endpoint: NextApiHandler = async (
...
+
+          if (response.status == StatusCodes.OK) {
+            res.send({
+              message: 'Export file generated successfully',
+            });
+          } else {
+            res.send({
+              message: 'Unable to generate export file',
+            });
+          }
```
Duslerke: It would help to log this. If it's a 400, you know to fix the request. If it's a 401 or 403 - you know it's credentials or perms issue, if it's 500, you know it's something beyond this app's control, etc.

pages/api/reports/internal/download.tsx
```diff
@@ -76,19 +76,21 @@ const endpoint: NextApiHandler = async (
...
       }
       break;
 
     default:
       res
-        .status(StatusCodes.BAD_REQUEST)
-        .json({ message: 'Invalid request method' });
+        .setHeader('Allow', 'POST')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
```
Duslerke: Only 1 legitimate case in this switch. This can be replaced with guard clause to drop the switch

pages/api/reports/internal/download.tsx
```diff
@@ -76,19 +76,21 @@ const endpoint: NextApiHandler = async (
             });
           }
         } catch (error) {
+          // Previously the raw Axios response and error were included in the
+          // response body, leaking internal details to the client.
+          console.error('Unable to download report', error);
```
Duslerke: This comment is about untouched code above.

Image

Lines 54 and 72:
Can line 54 condition ever be false? I'm thinking. Doesn't axios response always return a non-empty object with data / status / request fields? Meaning line 72 block cannot trigger. Perhaps, ln 54 condition should be `fileResponse.data` or something of the sort?

pages/api/auth/verify.tsx
```diff
@@ -92,8 +92,9 @@ const endpoint: NextApiHandler = async (
 
     default:
       res
-        .status(StatusCodes.BAD_REQUEST)
-        .json({ message: 'Invalid request method' });
+        .setHeader('Allow', 'POST')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
```
Duslerke: Only 1 case, switch can be removed

pages/api/auth/generate.tsx
```diff
@@ -26,8 +26,9 @@ const endpoint: NextApiHandler = async (
 ) => {
   if (req.method !== 'POST') {
     res
-      .status(StatusCodes.BAD_REQUEST)
-      .json({ message: 'Invalid request method' });
+      .setHeader('Allow', 'POST')
+      .status(StatusCodes.METHOD_NOT_ALLOWED)
+      .json({ message: 'Method not allowed' });
```
Duslerke, line 31: There we go: this is exactly what I was saying.
This makes code easier to read when this is within the guard cond instead.

pages/api/auth/exit.tsx
```diff
@@ -25,8 +25,9 @@ const endpoint: NextApiHandler = async (
 
     default:
       res
-        .status(StatusCodes.BAD_REQUEST)
-        .json({ message: 'Invalid request method' });
+        .setHeader('Allow', 'POST')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
```
Duslerke: Only 1 case, switch can be removed

pages/api/applications/index.ts
```diff
@@ -46,23 +46,33 @@ const endpoint: NextApiHandler = async (
...
-          res
-            .status(StatusCodes.FORBIDDEN)
-            .json({ message: 'Unable to add application' });
-          break;
-        }
+    case 'POST': {
+      if (!hasStaffPermissions(req)) {
+        res
+          .status(StatusCodes.FORBIDDEN)
+          .json({ message: 'Unable to add application' });
```
Duslerke: Since we're touching this code again, may as well replace the error message.
It should indicate that it's a permissions issue, and not a backend one.

pages/api/applications/index.ts
```diff
@@ -46,23 +46,33 @@ const endpoint: NextApiHandler = async (
...
-            .status(StatusCodes.FORBIDDEN)
-            .json({ message: 'Unable to add application' });
-          break;
-        }
+      if (hasReadOnlyStaffPermissions(req)) {
+        res
+          .status(StatusCodes.FORBIDDEN)
+          .json({ message: 'Unable to add application' });
+        break;
+      }
```
Duslerke: This block does the same thing as the block above.
So conditions can be combined:
`!hasStaffPermissions(req) || hasReadOnlyStaffPermissions(req)`

pages/api/applications/[id]/note.tsx
```diff
@@ -9,31 +11,48 @@ const endpoint: NextApiHandler = async (
...
       break;
+    }
 
     default:
       res
-        .status(StatusCodes.BAD_REQUEST)
-        .json({ message: 'Invalid request method' });
+        .setHeader('Allow', 'POST')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
```
Comment (Duslerke) [inline pages/api/applications/[id]/note.tsx]: Only 1 case in switch, so can remove switch to make code easier to read

pages/api/applications/[id]/note.tsx
```diff
@@ -9,31 +11,48 @@ const endpoint: NextApiHandler = async (
...
+        request = JSON.parse(req.body);
+      } catch (error) {
+        console.error('Unable to parse note request body', error);
+        res
+          .status(StatusCodes.BAD_REQUEST)
+          .json({ message: 'Unable to parse request' });
+        break;
+      }
+
+      const id = req.query.id as string;
```
Comment (Duslerke) [inline pages/api/applications/[id]/note.tsx]: Is `query` key guaranteed?
Thinking about the what if no query string is provided - will it crash due to trying to access `id` key from `undefined`.

pages/api/applications/[id]/note.tsx
```diff
@@ -9,31 +11,48 @@ const endpoint: NextApiHandler = async (
...
+
+      if (!canUpdateApplication(req, id)) {
+        res
+          .status(StatusCodes.FORBIDDEN)
+          .json({ message: 'Unable to add note' });
```
Comment (Duslerke) [inline pages/api/applications/[id]/note.tsx]: Error message should reflect that it's a permissions issue.
The way it's phrased now seems like it could have been a backend failure.

pages/api/applications/[id]/index.tsx
```diff
@@ -35,30 +44,27 @@ const endpoint: NextApiHandler = async (
...
-            // Request made and server responded
-            res.status(error.response.status).json(error.response.data);
-          } else if (error.request) {
-            // The request was made but no response was received
-            console.log(error.request);
-          } else {
-            // Something happened in setting up the request that triggered an Error
-            console.log('Error', error.message);
-          }
+      } catch (error) {
```
Comment (Duslerke) [inline pages/api/applications/[id]/index.tsx]: Line above (ln45) error message should say it's a permissions issue.

pages/api/applications/[id]/index.tsx
```diff
@@ -35,30 +44,27 @@ const endpoint: NextApiHandler = async (
...
       break;
+    }
 
     default:
       res
-        .status(StatusCodes.BAD_REQUEST)
-        .json({ message: 'Invalid request method' });
+        .setHeader('Allow', 'PATCH')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
```
Comment (Duslerke) [inline pages/api/applications/[id]/index.tsx]: Only 1 case in switch. Switch can be dropped.

pages/api/applications/[id]/index.tsx
```diff
@@ -18,10 +16,21 @@ const endpoint: NextApiHandler = async (
...
+        application = JSON.parse(req.body);
+      } catch (error) {
+        console.error('Unable to parse application request body', error);
+        res
+          .status(StatusCodes.BAD_REQUEST)
+          .json({ message: 'Unable to parse request' });
+        break;
+      }
+
+      const id = req.query.id as string;
```
Comment (Duslerke) [inline pages/api/applications/[id]/index.tsx]: Can query be `undefined`?

pages/api/applications/[id]/evidence.tsx
```diff
@@ -9,31 +11,48 @@ const endpoint: NextApiHandler = async (
...
+          .json({ message: 'Unable to parse request' });
+        break;
+      }
+
+      const id = req.query.id as string;
+
+      if (!canUpdateApplication(req, id)) {
+        res
+          .status(StatusCodes.FORBIDDEN)
+          .json({ message: 'Unable to update application' });
```
Comment (Duslerke) [inline pages/api/applications/[id]/evidence.tsx]: Error msg should indicate its a permissions issue.

pages/api/applications/[id]/evidence.tsx
```diff
@@ -9,31 +11,48 @@ const endpoint: NextApiHandler = async (
...
       break;
+    }
 
     default:
       res
-        .status(StatusCodes.BAD_REQUEST)
-        .json({ message: 'Invalid request method' });
+        .setHeader('Allow', 'POST')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
```
Comment (Duslerke) [inline pages/api/applications/[id]/evidence.tsx]: Only 1 case in switch, switch can be dropped.

pages/api/applications/[id]/complete.tsx
```diff
@@ -22,17 +23,22 @@ const endpoint: NextApiHandler = async (
...
       }
       break;
 
     default:
       res
-        .status(StatusCodes.BAD_REQUEST)
-        .json({ message: 'Invalid request method' });
+        .setHeader('Allow', 'PATCH')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
```
Comment (Duslerke) [inline pages/api/applications/[id]/complete.tsx]: Only 1 case in switch, switch can be dropped and this moved to guard block

pages/api/applications/[id]/complete.tsx
```diff
@@ -22,17 +23,22 @@ const endpoint: NextApiHandler = async (
             .json({ message: 'Unable to update application' });
         }
       } catch (error) {
-        console.error(error);
-        res
-          .status(StatusCodes.INTERNAL_SERVER_ERROR)
-          .json({ message: 'Unable to update application' });
+        if (axios.isAxiosError(error) && error.response) {
```
Comment (Duslerke) [inline pages/api/applications/[id]/complete.tsx]: Line 23 err msg does not tell its perms issue

pages/api/admin/logout.tsx
```diff
@@ -24,6 +24,12 @@ const endpoint: NextApiHandler = async (
           .json({ message: 'Unable to sign out' });
       }
       break;
+
+    default:
+      res
+        .setHeader('Allow', 'GET')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
```
Duslerke: Can be moved to guard block, which will allow dropping the entire switch indent.

pages/api/address/[postcode]/index.ts
```diff
@@ -27,8 +27,9 @@ const endpoint: NextApiHandler = async (
 
     default:
       res
-        .status(StatusCodes.BAD_REQUEST)
-        .json({ message: 'Invalid request method' });
+        .setHeader('Allow', 'GET')
+        .status(StatusCodes.METHOD_NOT_ALLOWED)
+        .json({ message: 'Method not allowed' });
```
Comment (Duslerke) [inline pages/api/address/[postcode]/index.ts]: Can be moved to a guard block. Then switch be dropped.

lib/gateways/notify-api.ts
```diff
@@ -6,55 +6,51 @@ import { NotifyClient } from 'notifications-node-client';
...
-  const notifyClient = new NotifyClient(process.env.NOTIFY_API_KEY);
-  const response = await notifyClient
-    .sendEmail(process.env.NOTIFY_TEMPLATE_MEDICAL_NEED, request.emailAddress, {
-      personalisation: request.personalisation,
-      reference: request.reference,
-    })
-    .then((response: NotifyResponse) => console.log(response))
-    .catch((err: NotifyResponse) => captureNotifyError(err, request));
+): Promise<NotifyResponse> =>
+  sendEmail(process.env.NOTIFY_TEMPLATE_NEW_APPLICATION, request);
```
Duslerke, line 46: I'm guessing the `sendEmail` and other moethods here were modified to where it extracts `personalisation` and `references` from the `request` by itself?

pages/api/reports/novalet/generate.tsx
sam-drumm: Yeah Ok - will update for routes with single method

lib/gateways/notify-api.ts
sam-drumm: Yeah that's right - covered off in #564

pages/api/applications/[id]/note.tsx
Comment (sam-drumm) [reply pages/api/applications/[id]/note.tsx]: don't think it's possible - could guard it but with Next it'll always be an object on routes.

pages/api/applications/index.ts
```diff
@@ -72,18 +61,18 @@ const endpoint: NextApiHandler = async (
         break;
       }
 
+      if (!hasStaffPermissions(req) || hasReadOnlyStaffPermissions(req)) {
+        res.status(StatusCodes.FORBIDDEN).json({
+          message: isStaffAction(application)
+            ? 'Unable to add application with assessment'
+            : 'Unable to add application',
```
Duslerke, line 68: Should likely say `Unauthorized to add...`

Duslerke (approved):
Looks good!

